PR #2, opened 2021-03-15
Backend registry

Speierers:
This PR seperated the pointer registry datastructures for the LLVM and CUDA backends.

### DIFF
--- a/src/registry.cpp
+++ b/src/registry.cpp
@@ -13,17 +13,19 @@
 static_assert(sizeof(void*) == 8, "32 bit architectures are not supported!");
 
 /// Register a pointer with Enoki's pointer registry
-uint32_t jitc_registry_put(const char *domain, void *ptr) {
+uint32_t jitc_registry_put(JitBackend backend, const char *domain, void *ptr) {
     if (unlikely(ptr == nullptr))
         jitc_raise("jit_registry_put(): cannot register the null pointer!");
 
+    Registry* registry = state.registry(backend);
+
     // Create the rev. map. first and throw if the pointer is already registered
-    auto it_rev = state.registry_rev.try_emplace(ptr, RegistryKey(domain, 0));
+    auto it_rev = registry->rev.try_emplace(ptr, RegistryKey(domain, 0));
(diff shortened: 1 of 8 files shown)


src/registry.cpp
```diff
@@ -251,38 +284,52 @@ void jitc_registry_set_attr(void *ptr, const char *name,
...
-            CUresult ret = cuMemAllocManaged((CUdeviceptr *) &ptr, new_size, CU_MEM_ATTACH_GLOBAL);
+            CUresult ret = cuMemAlloc((CUdeviceptr *) &ptr, new_size);
             if (ret != CUDA_SUCCESS) {
                 jitc_malloc_trim(true, true);
-                cuda_check(cuMemAllocManaged((CUdeviceptr *) &ptr, new_size, CU_MEM_ATTACH_GLOBAL));
+                cuda_check(cuMemAlloc((CUdeviceptr *) &ptr, new_size));
             }
-            cuda_check(cuMemAdvise((CUdeviceptr) ptr, new_size, CU_MEM_ADVISE_SET_READ_MOSTLY, 0));
+
+            // cuda_check(cuCtxSynchronize());
```
wjakob, line 295: This can be removed now?

src/vcall.cpp
```diff
@@ -246,7 +246,7 @@ void jitc_var_vcall(const char *name, uint32_t self, uint32_t mask,
 
         for (uint32_t j = se_offset[i]; j != se_offset[i + 1]; ++j)
             jitc_var_vcall_collect_data(vcall->data_map, data_size, i,
-                                        ts->side_effects[j]);
+                                        vcall->se[j - se_offset[0]]);
```
wjakob, line 249: Should this go into a separate commit?

wjakob (approved):
Looks great, thanks -- only two minor comments.

src/vcall.cpp
Speierers: yes!

src/registry.cpp
Speierers: Haven't tried this on Windows yet. But it should be yes

PR #3, opened 2021-03-30
Add double precision support to jitc_embree_trace

Speierers:
This PR adds support to double precision I/O to the `jitc_embree_trace` API. This is used in the parent `Mitsuba` project for enabling double precision ray tracing using the native KD-tree.


wjakob:
Looks great! :shipit:

PR #18, opened 2022-09-02
Optix pipeline and SBT management refactoring

Speierers:
## Overview

This PR addresses long standing issues with the management of Optix pipelines and SBT in Dr.Jit:

- OptiX pipeline and SBT lifetime was left to the caller's responsability, although it couldn't be aware of pending raytracing operations in un-evaluated kernels.

- OptiX pipeline and SBT was configured per thread, stored in the `ThreadState`, making it impossible to evaluate a kernel containing ray tracing operations traced in another thread.

- `jit_optix_configure()` had to be called before every call to `jit_optix_ray_trace()` to copy the pipeline and SBT data. This caused unecessary `memcpy` and didn't work reliably when multiple pipelines and SBT would be used simultaneously on the same thread.

## Solution

This patch associated an Optix pipeline and an Optix SBT to every ray tracing operation. To achieve this, we create two special JIT variables responsible for the management of those resources:
- `jit_optix_configure_pipeline()` creates a JIT variable responsible for the `OptixModule` and program groups. 
- `jit_optix_configure_sbt()` creates a JIT variable responsible for the SBT.

The special JIT variable created in `jit_optix_ray_trace()` will now depend on those two JIT variables, ensure the lifetime of those resources goes beyond the one of the ray tracing operations. The creator of those variables should also keep a reference of those on his own.

Those two variables define custom callbacks for `extra.assemble` and `extra.callback` in order to properly setup the pipeline and SBT when assembling the kernel, and release the resources when the variable is freed.

### DIFF
--- a/include/drjit-core/optix.h
+++ b/include/drjit-core/optix.h
@@ -43,14 +43,30 @@ extern JIT_EXPORT void jit_optix_check_impl(int errval, const char *file,
  *
  * The returned index should be passed as argument to subsequent calls to
  * \c jit_optix_ray_trace in order to use this pipeline for the ray tracing
- * operations.
+ * operations. See the docstring of \c jit_optix_ray_trace for a small example
+ * of how those functions relate to each other.
  */
 extern JIT_EXPORT uint32_t
-jit_optix_configure(const OptixPipelineCompileOptions *pco,
-                    OptixModule module,
-                    const OptixShaderBindingTable *sbt,
-                    const OptixProgramGroup *pg,
-                    uint32_t pg_count);
+jit_optix_configure_pipeline(const OptixPipelineCompileOptions *pco,
+                             OptixModule module,
+                             const OptixProgramGroup *pg,
+                             uint32_t pg_count);
+
+/**
+ * \brief Inform Dr.Jit about an OptiX Shader Binding Table
+ *
+ * This function creates a JIT variable responsible for the lifetime management
+ * of the OptiX Shader Binding Table and returns its corresponding index. Once
+ * the reference count of this variable reaches zero, the OptiX resources
+ * related to this Shader Binding Table will be freed.
+ *
+ * The returned index should be passed as argument to subsequent calls to
+ * \c jit_optix_ray_trace in order to use this Shader Binding Table for the ray
+ * tracing operations. See the docstring of \c jit_optix_ray_trace for a small
+ * example of how those functions relate to each other.
+ */
+extern JIT_EXPORT uint32_t
+jit_optix_configure_sbt(const OptixShaderBindingTable *sbt, uint32_t pipeline);
 
 /**
   * \brief Insert a function call to optixTrace into the program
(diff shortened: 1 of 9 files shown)


src/optix_api.cpp
```diff
@@ -461,32 +460,112 @@ void *jitc_optix_lookup(const char *name) {
...
+        p->program_groups.push_back(pg[i]);
+    extra.callback_data = p;
+
+    // Set the OptiX pipeline to use when assembling the kernel
+    extra.assemble = [](const Variable */*v*/, const Extra &extra) {
+        ThreadState *ts = thread_state(JitBackend::CUDA);
+        OptixPipelineData *p = (OptixPipelineData*) extra.callback_data;
+        if (ts->optix_pipeline == state.optix_default_pipeline) {
+            jitc_log(InfoSym, "jit_assemble_cuda(): set OptiX pipeline for next kernel launch");
+            ts->optix_pipeline = p;
```
wjakob, line 495: neat trick!

include/drjit-core/optix.h
```diff
@@ -30,14 +31,40 @@ extern JIT_EXPORT void *jit_optix_lookup(const char *name);
...
-/// Inform Dr.Jit about a partially created OptiX pipeline
-extern JIT_EXPORT void
-jit_optix_configure(const OptixPipelineCompileOptions *pco,
-                    const OptixShaderBindingTable *sbt,
-                    const OptixProgramGroup *pg,
-                    uint32_t pg_count);
+/**
+ * \brief Inform Dr.Jit about a partially created OptiX pipeline
+ *
+ * This function creates a JIT variable responsible for the lifetime management
```
wjakob, line 39: Would you mind adding a small example of how `jit_optix_configure_pipeline` and `jit_optix_configure_sbt` relate to each other (otherwise users will have to dig into the Mitsuba code to understand how they are used).

src/internal.h
```diff
@@ -384,21 +390,9 @@ struct ThreadState {
...
-    OptixShaderBindingTable optix_shader_binding_table {};
-
-    /// User-provided list of program groups
-    std::vector<OptixProgramGroup> optix_program_groups;
-
-    /// Components for a tiny self-contained OptiX pipeline for testcases etc.
-    OptixProgramGroup optix_program_group_base = nullptr;
-    OptixModule optix_module_base = nullptr;
-    void *optix_miss_record_base = nullptr;
+    /// OptiX pipeline associated to the next kernel launch
```
wjakob, line 393: associated *with*

wjakob:
Looks great, just a few minor things.

wjakob (approved):
Looks great, just a few minor things.

src/init.cpp
```diff
@@ -258,6 +258,12 @@ void jitc_shutdown(int light) {
         }
     }
 
+#if defined(DRJIT_ENABLE_OPTIX)
+        // Free the default OptiX shader binding table and pipeline (ref counting)
+        if (state.optix_default_sbt_index)
+            jitc_var_dec_ref_ext(state.optix_default_sbt_index);
```
njroussel, line 264:
```suggestion
    // Free the default OptiX shader binding table and pipeline (ref counting)
    if (state.optix_default_sbt_index)
        jitc_var_dec_ref_ext(state.optix_default_sbt_index);
```

src/optix_api.cpp
```diff
@@ -516,10 +595,22 @@ bool jitc_optix_compile(ThreadState *ts, const char *buf, size_t buf_size,
...
+    int rv = optixModuleCreateFromPTX(
+        optix_context, &mco, &pipeline.compile_options, buf,
+        buf_size, error_log, &log_size, &kernel.optix.mod);
+    if (rv) {
+        jitc_log(Error, "jit_optix_compile(): optixModuleCreateFromPTX() "
+                 "failed. Please see the PTX assembly listing and error "
+                 "message below:\n\n%s\n\n%s", buf, error_log);
+        jitc_optix_check(rv);
+    }
+#else
```
njroussel, line 610: We can remove this, right ?

njroussel (approved):
:rocket: Looks good

PR #21, opened 2022-10-26
Fix raytracing operation in virtual function calls

Speierers:
This PR attempts to fix raytracing operation in virtual function calls and add the ability to ray trace multiple scenes within a single OptiX megakernel. This implies that those scenes share the same SBT, which requires the implementation of `jit_optix_update_sbt()`. 

This PR also disables the OptiX validation mode by default as it isn't reliable in current OptiX versions.

### DIFF
--- a/include/drjit-core/optix.h
+++ b/include/drjit-core/optix.h
@@ -69,6 +69,18 @@ extern JIT_EXPORT uint32_t
 jit_optix_configure_sbt(const OptixShaderBindingTable *sbt, uint32_t pipeline);
 
 /**
+ * \brief  Update existing OptiX Shader Binding Table data
+ *
+ * This function updates the Shader Binding Table data held by the JIT
+ * variable \c index previously created using \c jit_optix_configure_sbt. This
+ * update is necessary when adding more geometry to an existing scene or when
+ * sharing the OptiX pipeline and SBT across multiple scenes (e.g. ray tracing
+ * against different scenes within the same megakernel).
+ */
+extern JIT_EXPORT void
+jit_optix_update_sbt(uint32_t index, const OptixShaderBindingTable *sbt);
+
+/**
   * \brief Insert a function call to optixTrace into the program
   *
   * The \c args list should contain a list of variable indices corresponding to
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -185,7 +185,6 @@ uint32_t jit_record_checkpoint(JitBackend backend) {
     if (jit_flag(JitFlag::Recording))
         result |= 0x80000000u;
     return result;
-
 }
 
 uint32_t jit_record_begin(JitBackend backend) {
@@ -871,6 +870,11 @@ uint32_t jit_optix_configure_sbt(const OptixShaderBindingTable *sbt, uint32_t pi
     return jitc_optix_configure_sbt(sbt, pipeline);
 }
 
+void jit_optix_update_sbt(uint32_t index, const OptixShaderBindingTable *sbt) {
+    lock_guard guard(state.lock);
+    jitc_optix_update_sbt(index, sbt);
+}
+
 void jit_optix_ray_trace(uint32_t nargs, uint32_t *args, uint32_t mask,
                          uint32_t pipeline, uint32_t sbt) {
     lock_guard guard(state.lock);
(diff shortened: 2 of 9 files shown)


include/drjit-core/optix.h
```diff
@@ -68,6 +68,15 @@ jit_optix_configure_pipeline(const OptixPipelineCompileOptions *pco,
 extern JIT_EXPORT uint32_t
 jit_optix_configure_sbt(const OptixShaderBindingTable *sbt, uint32_t pipeline);
 
+/**
+ * \brief  Update existing OptiX Shader Binding Table data
+ *
+ * This function updates the Shader Binding Table data hold by the JIT
+ * variable \c index previously created using \c jit_optix_configure_sbt.
```
wjakob, line 75: Could you add a line saying what parts of the SBT updated, and in what context such an update becomes necessary?

src/api.cpp
```diff
@@ -199,7 +201,8 @@ void jit_record_end(JitBackend backend, uint32_t value) {
 
     // Set recording flag to previous value
     jit_set_flag(JitFlag::Recording, value & 0x80000000u);
-    value &= 0x7fffffff;
+    thread_state(backend)->recording_vcall = value & 0x40000000u;
```
wjakob, line 204: Is this representation (`recording_vcall`) sufficiently general? IMO we can have arbitrary levels of loops and vcalls. It would seem to me that we need some stack, or that the logic in `state.h` stores the previous value and updates it similar to other state flags in there.

src/eval_llvm.cpp
```diff
@@ -612,14 +612,18 @@ void jitc_llvm_ray_trace(uint32_t func, uint32_t scene, int occluded,
...
+
     jitc_log(InfoSym, "jitc_llvm_ray_trace(): tracing %u %sray%s%s%s", size,
              occluded ? "shadow " : "", size != 1 ? "s" : "",
              placeholder ? " (part of a recorded computation)" : "",
              double_precision ? " (double precision)" : "");
 
     Ref op = steal(jitc_var_new_stmt_n(JitBackend::LLVM, VarType::Void,
-                               occluded ? "// Ray trace (occluded)"
-                                        : "// Ray trace",
+                               within_vcall ? (occluded ? "// Ray trace (occluded) (within_vcall)"
```
wjakob, line 623: can remove the ``_`` in the comment.

wjakob: Since you are making the change, would you mind renaming `occluded` to `shadow ray` (and `shadow_ray` in the parameter)? Occluded sounds weird to me, because it sounds like an occlusion has already been found.

src/eval_llvm.cpp
```diff
@@ -647,7 +651,8 @@ void jitc_llvm_ray_trace(uint32_t func, uint32_t scene, int occluded,
 static void jitc_llvm_ray_trace_assemble(const Variable *v, const Extra &extra) {
     const uint32_t width = jitc_llvm_vector_width;
     const uint32_t id = v->reg_index;
-    bool occluded = strstr(v->stmt, "occluded") != nullptr;
+    bool occluded  = strstr(v->stmt, "occluded") != nullptr;
```
wjakob, line 654: change needs to be mirrored here.

src/api.cpp
wjakob: NVM, I will suggest a different approach in the main review comment.

src/vcall.cpp
```diff
@@ -288,12 +289,12 @@ uint32_t jitc_var_vcall(const char *name, uint32_t self, uint32_t mask,
         for (uint32_t j = 0; j < n_out; ++j)
             jitc_var_vcall_collect_data(vcall->data_map, data_size, i,
                                         out_nested[j + i * n_out],
-                                        vcall->use_self);
```
njroussel, line 291: Shouldn't we have a logical or on the `optix` value here ?
In case we already have it set to `true`, but the `jitc_var_vcall_collect_data` writes `false`.

src/vcall.cpp
```diff
@@ -1329,6 +1331,9 @@ void jitc_var_vcall_collect_data(tsl::robin_map<uint64_t, uint32_t, UInt64Hasher
     if (!v->literal && v->stmt && strstr(v->stmt, "self"))
         use_self = true;
 
+    if (v->optix)
+        optix = true;
```
njroussel, line 1335: Ignore my previous comment, I didn't see that it could only write `true`

include/drjit-core/optix.h
```diff
@@ -68,6 +68,18 @@ jit_optix_configure_pipeline(const OptixPipelineCompileOptions *pco,
 extern JIT_EXPORT uint32_t
 jit_optix_configure_sbt(const OptixShaderBindingTable *sbt, uint32_t pipeline);
 
+/**
+ * \brief  Update existing OptiX Shader Binding Table data
+ *
+ * This function updates the Shader Binding Table data hold by the JIT
```
wjakob, line 74: `hold` -> `held`

wjakob (commented):
The change looks reasonable to me, only minor change requests to me. 

However: before committing it, could you please test that it works as expected? This would involve recording a virtual function call that actually performs two separate ray tracing steps to different scenes, and checking that the output is as expected. Perhaps you have already done this, but it wasn't clear to me from the PR.

PR #22, opened 2022-11-01
Automatic evaluation of external scalar literals in recorded vcalls

Speierers:
To reduce the need for `dr::opaque` and `dr::make_opaque` in parent libraries (e.g. Dr.Jit and Mitsuba 3), this patch aims at automatically detecting scalar literal variables that are created outside of the recorded virtual functions. Those variables then receive a special treatment:

- Evaluated in-place (similar to what `dr::make_opaque` would do).
- Disable constant propagation for all variables depending on this external scalar literal variable.

To detect whether a scalar literal variable was created outside of the virtual function, this patch relies on the `v->cse_scope` value as Dr.Jit always instantiates a new CSE scope when tracing virtual functions. So it is enough to check whether the current CSE scope is different from the one of the used variable.


src/op.cpp
```diff
@@ -452,22 +452,46 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
...
     bool const_prop = jitc_flags() & (uint32_t) JitFlag::ConstProp;
 
     if (unlikely(n_dep == 0 || n_dep > 4))
         jitc_fail("jit_var_new_op(): 1-4 dependent variables supported!");
 
+    uint32_t backend_i = 0;
     for (uint32_t i = 0; i < n_dep; ++i) {
         if (likely(dep[i])) {
             Variable *vi = jitc_var(dep[i]);
+            backend_i |= (uint32_t) vi->backend;
```
wjakob, line 465: The `|=` seems odd to me. I would just do an assignment.

src/op.cpp
```diff
@@ -452,22 +452,46 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
...
+        }
+    }
+
+    JitBackend backend = (JitBackend) backend_i;
+
+    /* When recording a virtual function call, disable constant propagation for
+       all variables that depend on a scalar literal that was created outside of
+       this virtual function call as this variable will later be evaluated to
+       avoid baking it into the kernel IR. */
+    ThreadState *ts = thread_state(backend);
```
wjakob, line 478: That's a good point about disabling constant propagation 👍

src/vcall.cpp
```diff
@@ -1326,11 +1326,18 @@ void jitc_var_vcall_collect_data(tsl::robin_map<uint64_t, uint32_t, UInt64Hasher
...
 
-    const Variable *v = jitc_var(index);
+    Variable *v = jitc_var(index);
 
     if (!v->literal && v->stmt && strstr(v->stmt, "self"))
         use_self = true;
 
+    /* Evaluate scalar literals created outside of this virtual function call
+       to avoid baking them inside of the kernel IR. */
+    if (v->literal && v->cse_scope != thread_state(v->backend)->cse_scope) {
```
wjakob, line 1336: What I had in mind is that -- rather than actually evaluating those literals, we can handle them specially in the part of the code that generates the data array used by the virtual function call. I am referring to the loop at  https://github.com/mitsuba-renderer/drjit-core/blob/master/src/vcall.cpp#L336.

This loop computes the input that is later used to launch a kernel which either

1. Copies a raw pointer value (which is actually a kind of literal) from the CPU into a GPU output array
2. Copies a single data word (1, 2, 4, or 8 bytes) from some region of GPU memory into a GPU output array

What I am suggesting is that we have a 3rd case, which copies any other kind of constant literal (1, 2, 4, or 8 bytes) form CPU memory to the GPU output array. So this would be a sort of generalization of #1.

Speierers:
@wjakob this is ready for another round of review. I basically applied your comment on how to implement this in a different way.

resources/Makefile
```diff
@@ -1,5 +1,5 @@
 COMPUTE_CAPABILITY=compute_70
-CUDA_VER=10.2
+CUDA_VER=11.8
```
wjakob, line 2: It seems unlikely that we need a newer cuda versions and a higher compute compability for these changes (-> huge diff). Would you mind recompiling with the old 10.2 nvcc?

resources/misc.cuh
```diff
@@ -35,6 +35,7 @@ struct VCallDataRecord {
     uint32_t offset;
     uint32_t size;
     const void *src;
```
wjakob, line 37: Can we rename `src` into `value` given that it takes on a new identity as literal storage location?

src/eval_cuda.cpp
```diff
@@ -321,31 +321,37 @@ void jitc_assemble_cuda_func(const char *name, uint32_t inst_id,
                            "    setp.ne.u16 %%p%u, %%w0, 0;\n",
                            v->param_offset, v->reg_index);
             }
-        } else if (v->data || vt == VarType::Pointer) {
+        } else if (v->literal || v->data) {
```
wjakob, line 324: I am confused: do we want to take this branch this for all literals? I would assume that most literals still go through the ``jitc_render_stmt_cuda`` or ``jitc_render_stmt_llvm`` functions. It's only the literals that were created before recording the vcall that should be opaque-ified in this way. Maybe I am misunderstanding.

src/op.cpp
```diff
@@ -452,44 +452,27 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
...
+        uint32_t idx = dep[i];
+        if (likely(idx)) {
             Variable *vi = jitc_var(dep[i]);
             vti = std::max(vti, vi->type);
             size = std::max(size, vi->size);
             dirty |= (bool) vi->ref_count_se;
             placeholder |= (bool) vi->placeholder;
-            backend_i |= (uint32_t) vi->backend;
             v[i] = vi;
-
```
wjakob, line 471: Curious about this removal.

src/vcall.cpp
```diff
@@ -1326,7 +1326,12 @@ void jitc_var_vcall_collect_data(tsl::robin_map<uint64_t, uint32_t, UInt64Hasher
     if (!it_and_status.second)
         return;
 
-    const Variable *v = jitc_var(index);
+    Variable *v = jitc_var(index);
+    ThreadState *ts = thread_state(v->backend);
+
+    /* Scalar literals created outside of this virtual function call should be
+       considered as data to avoid baking them inside of the kernel IR. */
+    bool ext_literal = v->literal && v->cse_scope != ts->cse_scope;
```
wjakob, line 1334: We also change CSE scope in other situations (e.g. loop recording). Maybe this is too broad? Perhaps it's better store an index somewhere that records the location of the innermost function being recorded right now. Anything older than that is opaque-ified.

wjakob (commented):
Great, looks like we are on the same page. A few more comments for round two :).

src/op.cpp
Speierers: This was moved down as we first need to check whether to use constant propagation or not.

src/eval_cuda.cpp
Speierers: We indeed take this branch for all literals. At this point we can't use the heuristic based on CSE scope (or now indices order) to determine whether this branch should be taken or not. But we then check the offset in the datamap, which will be `-1` for all literal values that can be baked in the kernel.

src/vcall.cpp
Speierers: I am now using the `ts->vcall_self_index` variable which is created right before tracing the virtual function, hence it can act as a lower bound for the variable index to detect external scalar literals in vcalls.

wjakob: great!

resources/misc.cuh
```diff
@@ -44,15 +47,18 @@ KERNEL void vcall_prepare(void *out, const VCallDataRecord *rec_, uint32_t size)
...
     void *dst = (uint8_t *) out + rec.offset;
 
-    switch (rec.size) {
-        case 0: *(uint64_t *) dst = (uint64_t)    src; break;
-        case 1: *(uint8_t *)  dst = *(uint8_t *)  src; break;
-        case 2: *(uint16_t *) dst = *(uint16_t *) src; break;
-        case 4: *(uint32_t *) dst = *(uint32_t *) src; break;
-        case 8: *(uint64_t *) dst = *(uint64_t *) src; break;
+    if (rec.literal) {
+        memcpy(dst, &rec.value, rec.size);
```
wjakob, line 54: `memcpy` can be a surprisingly expensive operation when it's not straightforward for the compiler to see how much there is to copy. In this case, `size` is only an 8-bit value (so a dynamic copy of 0..255 bytes), but I don't trust that the compiler is smart enough to figure it out. How about rewriting it as follows (merging the `literal` and non-`literal` case):

```
        switch (rec.size) {
             case 1: *(uint8_t *)  dst = literal ? (uint8_t)  value : *(uint8_t *)  value; break;
             case 2: *(uint16_t *) dst = literal ? (uint16_t) value : *(uint16_t *) value; break;
             case 4: *(uint32_t *) dst = literal ? (uint32_t) value : *(uint32_t *) value; break;
             case 8: *(uint64_t *) dst = literal ? (uint64_t) value : *(uint64_t *) value; break;
         }
         ```

File: src/vcall.cpp [hunk 10]
```diff
@@ -1326,7 +1326,12 @@ void jitc_var_vcall_collect_data(tsl::robin_map<uint64_t, uint32_t, UInt64Hasher
     if (!it_and_status.second)
         return;
 
-    const Variable *v = jitc_var(index);
+    Variable *v = jitc_var(index);
```
wjakob, line 1329: The logic here looks good to me.

wjakob (approved):
Perfect, this looks all good to me now. Did you test it with mitsuba and some of the `dr::make_opaque` entries removed?

Just one small comment about the CUDA kernel and `memcpy`.

PR #23, opened 2022-11-02
Fix jit_var(r0): unknown variable when recording virtual calls

tomix1024:
This is a fix for mitsuba-renderer/mitsuba3#356.

When a loop gets recorded and some of the dependent input/output variables are optimized out, the respective `indices` referenced in `Extra::dep` get replaced by `0` entries in [`jitc_var_loop()`](https://github.com/tomix1024/drjit-core/blob/044f04b54fce4af81fd841352fed594889e6b608/src/loop.cpp#L210).

When this loop is then part of a `jitc_var_vcall_collect_data()` traversal, a variable with an id (`index`) of `0` is referenced that does not exist and an exception is raised.

A simple check if the index of the dependent variable is zero before calling `jitc_var_vcall_collect_data()` recursively fixes the problem.


wjakob:
LGTM, thanks!

PR #24, opened 2022-11-09
Check device feature flags before using async memory allocations on CUDA

njroussel:
This PR fixes a memory allocation issue on older GPUs (issue [here](https://github.com/mitsuba-renderer/mitsuba3/issues/182)).

Asynchronous memory allocations were only added in CUDA 11.2,  and are not supported on all GPUs. A device attribute indicating whether this feature is supported exists and this PR makes use of it.


Speierers (commented):
Looks good to me.

wjakob:
Many thanks for tracking this down @njroussel. The change looks good to me.

PR #26, opened 2022-11-11
Expose new functions to manually push/pop CUDA contexts

njroussel:
This PR adds helpers to switch CUDA contexts.

For OptiX, we expose a `OptixDeviceContext` throught `jit_optix_context()`. This context depends on a CUDA context and if users want to re-use it for other OptiX computations outside of Dr.Jit they cannot do it without setting the same CUDA context.
These new helpers will allow users to do exactly this or, more generally, scope any other CUDA computations they might want to do without Dr.Jit.

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -200,6 +200,12 @@ extern JIT_EXPORT void* jit_cuda_stream();
 /// Return the CUDA context associated with the current thread
 extern JIT_EXPORT void* jit_cuda_context();
 
+/// Push a new CUDA context to be associated with the current thread
+extern JIT_EXPORT void jit_cuda_push_context(void *);
+
+/// Pop the CUDA context associated to the current thread and return it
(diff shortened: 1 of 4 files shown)


Speierers (commented):
Looks good to me 🚀

wjakob:
This is very anecdotal and my memory is a little foggy: Early in the development of Dr.Jit (then called Enoki), I tried to use this push/pop context feature and ran into issues with other tensor libraries (PyTorch or Tensorflow). Somehow popping the stack wasn't consistently returning things to the original condition, and that caused issues when intermingling Dr.Jit calls with those of the other framework. Hence the current raw way of setting/getting the context in Dr.Jit with the "stack" being maintained by a RAII wrapper.

njroussel:
Good to know, the way I am currently using these the new functions is Mitsuba is also through a new RAII wrapper (I couldn't expose the existing one because of the C interface)

PR #28, opened 2022-11-15
Disable constant propagation on operations involving the default mask

njroussel:
This PR disables constant propagation on operations involving the default mask. This fixes cases where we **need** to evaluate a JIT variable that happens also to be the default mask because of constant propagation. (follow-up to https://github.com/mitsuba-renderer/drjit-core/pull/25).

For example, in wavefront loops the condition must be evaluated on the first iteration. It often is the case that the loop condition is simply a literal evaluating to true at the first iteration. Hence the operation which mixes the condition and the mask ([here](https://github.com/mitsuba-renderer/drjit/blob/master/include/drjit/loop.h#L221)) will simply return the mask.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -459,6 +459,13 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
     if (unlikely(n_dep == 0 || n_dep > 4))
         jitc_fail("jit_var_new_op(): 1-4 dependent variables supported!");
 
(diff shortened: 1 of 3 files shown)


wjakob:
Ooops. I am slightly worried where that fix is taking us (lots of fixups related to default masks in different places).

Perhaps the right thing after all is to construct default masks of a certain size and allow them to be evaluated. Let me look into that..

wjakob:
Feel free to revert my previous commit in the meantime

njroussel:
Replaced by #32

PR #29, opened 2022-11-15
Auto opaque vcall  - 3rd attempt

Speierers:
Here is my 3rd attempt at implementing the handling of external literal in recorded virtual function calls.

The previous patch wasn't optimal as it would prevent the constant propagation of literal arguments through virtual function calls. While de-virtualization was working properly, it still meant that large data-structure arguments with a lot of zero-members would generate a lot of un-necessary "opaque" IR that would turn out to yield zeros.

This is especially important for automatic differentiation when only a subset of the input arguments will have non-zero gradients.

Looking back at the original purpose of those changes, we are only interested in "un-baking"  literal variables created before tracing the VCalls that are directly accessible within the functions (e.g. plugin parameters). With this in mind, it should be enough to ignore the input arguments.

**Moreover**, this patch also ignores `Bool` variables in this process. This is to avoid un-baking masks (e.g. disabling constant propagation) that are computed in between / after virtual function tracing. Also, it is very unlikely for a plugin parameter to use the `Mask` type, hence this seems like a reasonable solution.

### DIFF
--- a/src/eval_cuda.cpp
+++ b/src/eval_cuda.cpp
@@ -311,7 +311,7 @@ void jitc_assemble_cuda_func(const char *name, uint32_t inst_id,
             }
         }
 
-        if (v->vcall_iface) {
+        if (v->vcall_iface && !v->literal) {
             if (vt != VarType::Bool) {
                 buffer.fmt("    ld.param.%s %s%u, [params+%u];\n",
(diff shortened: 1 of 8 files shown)


njroussel:
LGTM :smiley: 

Doesn't look like the CI wants to pickup this PR, let me have a look

wjakob:
LGTM!

PR #31, opened 2022-11-16
Remove masking from hardware texture operations

njroussel:
The functions `jit_cuda_tex_lookup` and `jitc_cuda_tex_bilerp_fetch` would previously take a mask as an argument which was used to fill the result with zeros at the masked indices. This PR removes the `mask` arguments entirely.

The output variables for both functions are always new JIT variables, users can still zero-out certain indices if they need to with a second operation. Additionally, because there is always a wrapping feature on a texture, there will never be invalid lookup coordinates (NaNs and infinites are supported too).
Masking at this level was therefore unnecessary.

### DIFF
--- a/include/drjit-core/texture.h
+++ b/include/drjit-core/texture.h
@@ -102,7 +102,7 @@ extern JIT_EXPORT void jit_cuda_tex_memcpy_t2d(size_t ndim, const size_t *shape,
  */
 extern JIT_EXPORT void jit_cuda_tex_lookup(size_t ndim,
                                            const void *texture_handle,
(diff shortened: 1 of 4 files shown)


wjakob:
Looks great! Let's merge if it passes the tests.

wjakob:
For some reasons the CI is no longer auto-triggering for `drjit-core` PRs

njroussel:
Indeed, I noticed it this morning and ran all pending builds by hand hoping that one of them was blocking the rest. Clearly, there is something else going on :disappointed:

Speierers (commented):
LGTM 🚀

PR #33, opened 2022-11-17
Support large launches with up to 2^32 threads on OptiX backend

wjakob:
Mitsuba rendering tasks executed using Dr.Jit's OptiX backend are currently limited to 2^30 ≈ 1 billion Monte Carlo samples in total, which sounds like a lot. However, that's just enough to render a full HD image using 512 samples per pixel:

   1920*1080*512 = 1'061'683'200 < 1'073'741'824 = 2^32

This commit increases the limit to 4 billion, which corresponds to 2048 samples per pixel at 1080p resolution. This is the *absolute maximum* wavefront size that Dr.Jit can accommodate due to its 32 bit based indexing.

It's worth mentioning that further increasing that limit by upgrading to 64 bit indexing is not a good idea as this comes with significant costs on modern GPUs that lack 64 bit registers. Instead, multiple rendering passes should be used in the renderer to handle such very large rendering tasks.

What this commit commit enables is to internally run multiple passes without re-tracing when the sample count is between 2^30 and 2^32, which makes it less likely that users would require the `n_passes` parameter of the renderer (which arguably adds a few complications).

### DIFF
--- a/src/optix_api.cpp
+++ b/src/optix_api.cpp
@@ -6,6 +6,7 @@
 #include "eval.h"
 #include "var.h"
 #include "op.h"
+#include "util.h"
 
 #define OPTIX_ABI_VERSION 55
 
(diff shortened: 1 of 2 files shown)


wjakob:
The CI failure is an unrelated issue that's also broken on master. Likely that slipped through while the CI was not getting triggered. My other PR (#32) fixes it.

src/optix_api.cpp
```diff
@@ -810,17 +811,25 @@ void jitc_optix_launch(ThreadState *ts, const Kernel &kernel,
         sbt.callablesRecordCount = kernel.optix.pg_count - 1;
     }
 
-    if (launch_size >= 0x40000000u)
-        jitc_raise("jit_optix_launch(): attempted to launch a very large "
-                   "wavefront of size %u. The maximum wavefront size permitted "
-                   "by OptiX is 2**30 == 1073741824. Please render using "
-                   "multiple passes, use fewer samples, or a lower resolution.",
-                   launch_size);
```
Speierers, line 818: Don't we still need a error message when the wavefront size is greater than 2^32?

Speierers (commented):
Nice to be able to bump the wavefront size a bit for the OptiX backend!

src/optix_api.cpp
wjakob: You cannot create Dr.Jit arrays wit more than 2^32 entries. The `jit_var_new*` variables will generate an error message. That error message is arguably a little bit low level in the rendering context, and that's why the Mitsuba part also has a more detailed message relating this to the image resolution and number of spp.

njroussel (approved):

PR #34, opened 2022-11-19
Submit smaller work batches on Windows with the WDDM driver model

wjakob:
A curiosity of Windows graphics drivers for NVIDIA GPUs is that they can run in two different modes:

- WDDM mode: the default, which involves the windows display driver.

- TCM mode: only available on production cards like Quadro/Tesla, and when the card is in headless mode (no screen attached). The NVIDIA driver manages the card on its own without going through the windows display driver subsystem.

The main drawback of WDDM mode is that a long-running kernel freezes the UI for that amount of time--there is no multi-tasking. If the kernel runs for too long (more than a few seconds), it will eventually trigger a driver reset that crashes the kernel. On Linux, the graphics stack runs concurrently so that issue does not arise.

We did not have precautions for this in Dr.Jit so far, and the consequence is that running longer-running rendering jobs on Windows is likely painful. This commit builds on my previous change to OptiX kernel submission and sets a lower maximum number of threads (roughly 1 full HD frame at 1 sample per pixel) when the system detects a GPU running in WDDM mode on Windows.


Speierers (commented):
LGTM

PR #35, opened 2022-11-22
Remove PTX `@p` predicates

Speierers:
This patch gets rid of the use of `@pXXX ...` predicated instructions in the PTX generation for `dr.gather`, `dr.scatter` and `dr.select()`.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -1262,8 +1262,10 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
                     stmt = literal_zero[2] ? "selp.$b0 $r0, $r2, 0, $r1"
                                            : "selp.$t0 $r0, $r2, $r3, $r1";
                 } else {
-                    stmt = "@$r1 mov.pred $r0, $r2$n"
-                           "@!$r1 mov.pred $r0, $r3";
+                    // r0 = (r1 && r2) || (!r1 && r3)
+                    stmt = "and.pred %p3, $r1, $r2$n"
+                           "and.pred %p2, !$r1, $r3$n"
+                           "or.pred $r0, %p2, %p3";
                 }
             } else {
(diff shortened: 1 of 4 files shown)


wjakob:
This looks good to me. It would be good to benchmark it against some scenes to double-check that there are no perf. regressions.

wjakob:
I didn't see a change related to `dr.select` that the PR description mentions.

src/op.cpp
```diff
@@ -1287,8 +1287,12 @@ uint32_t jitc_var_new_op(JitOp op, uint32_t n_dep, const uint32_t *dep) {
...
                                            : "selp.$t0 $r0, $r2, $r3, $r1";
                 } else {
-                    stmt = "@$r1 mov.pred $r0, $r2$n"
-                           "@!$r1 mov.pred $r0, $r3";
+                    stmt = "@!$r1 bra l_$i0_masked$n"
+                           "mov.pred $r0, $r2$n"
+                           "bra.uni l_$i0_done;\n"
+                           "l_$i0_masked:\n    "
+                           "mov.pred $r0, $r3;\n"
+                           "l_$i0_done:";
```
Speierers, line 1295: @wjakob this is for the `dr.select` statement.

njroussel (approved):
Looks good :rocket:

src/op.cpp
wjakob: That seems a bit extreme to use jumps here. Can we express this using binary logic? Basically we need PTX instructions for the ternary operator (a ? b : c). For example: `(a && b) || (!a && c)`. Maybe there is a more efficient way.

Speierers: Something like this maybe?
``` 
                    // r0 = (r1 && r2) || (!r1 && r3)
                    stmt = "and.$b0 $r2, $r1, $r2$n"
                           "not.$b0 $r1, $r1$n"
                           "and.$b0 $r3, $r1, $r3$n"
                           "or.$b0 $r0, $r2, $r3$n";
```

PR #36, opened 2022-11-22
Revert automatic `make_opaque` of literals in virtual function calls

njroussel:
This PR reverts commit 7af0a7394ab07fb6b4dfe6107d4f66126282a353 and 0d9a1a3c02619cfade3c277817a06ce8bc26c2e5.

A few extra changes still need to be made for this feature to be shipped. These commits are therefore reverted until a future release.

(The reverts were squashed into a single commit, hopefully it will make it easier to re-introduce theses changes when we'll want to).


wjakob:
the test failure on Windows could be a flaky test indicating a low-probability failure mode. I think I've seen that once before.

PR #38, opened 2022-11-23
Merged internal and external reference count fields

wjakob:
The Variable data structure currently keeps track of an internal and an external reference count. We aren't really using that property, and it consumes precious storage in an otherwise highly optimized data structure. This experimental commit merges them into a joint 32 bit reference count field to free up memory for subsequent refactoring.


Speierers (commented):
It's difficult to truly review this PR (e.g. spot a missing change) but I suppose if this compile and passes the tests we are good to go 🚀

wjakob:
Could I ask you to do the `drjit` parts? (Should be a simple search+replace job) and then merge if it passes the tests?

njroussel (approved):
Looks good :100:

PR #39, opened 2022-11-27
Further variable data structure refactoring

wjakob:
This commit builds on PR #38 and further refactors the `Variable` data structure by adding a ``kind`` field that specifies the kind of the variable in question. The field must be set to one of:

- `VarKind::Literal`: constant literals
- `VarKind::Stmt`: a statement in an intermediate representation (LLVM IR, PTX)
- `VarKind::Data`: a pointer to device memory

This permits overlapping `data`, `literal`, and `stmt` using a union, which frees up 64 bits (`data` was previously stored separately):

```cpp
union {
    uint64_t literal;
    const char *stmt;
    void *data;
}
```

The `Variable` struct was extended with convenience functions to check `kind`:

```cpp
inline bool is_literal() { return kind == (uint32_t) VarKind::Literal; }
inline bool is_stmt()    { return kind == (uint32_t) VarKind::Stmt; }
inline bool is_data()    { return kind == (uint32_t) VarKind::Data; }
```

Most of the PR then consists of adding checks that call these helper functions to disambiguate the kind of variable being processed.

A few other minor changes are rolled in:

- The Variable data structure gets a helper function `is_dirty()` that wraps the somewhat opaque expression `ref_count_se > 0` previously used in various places.

- CSE is renamed to LVN (local value numbering) throughout the codebase. CVN and LVN are related but not quite the same, and LVN is what is implemented in Dr.Jit.

- The Variable `cse_scope` field was renamed to `scope` and extended to 32 bit. The name change is motivated by the more general use of this variable in the future. A similar renaming change was made to API, which affects the `jit_set_scope()`, `jit_scope()`, and `jit_new_scope()` functions in `jit.h`.

- The `ref_count_se` field was extended to 32 bits.

- There was still space left over in the `Variable` data structure, so a 32 bit variable named `unused` was added to the data structure to keep the size of a hash table entry at 64 bytes (1 cache line). This field will be default-initialized to zero.

- The second commit adds a test for the symbolic gather feature. `dr.gather()` has the ability to rewrite an input array if it is in symbolic form, but there wasn't a test to check that this rewriting is actually happening.

### DIFF
--- a/tests/basics.cpp
+++ b/tests/basics.cpp
@@ -70,11 +70,11 @@ TEST_BOTH(03_load_store_mask) {
         uint32_t mask = jit_var_new_op_2(JitOp::Eq, odd, zero);
 
         jit_assert(strcmp(jit_var_str(mask),
-                           i == 0 ? "[1]" : "[1, 0, 1, 0, 1, 0, 1, 0, 1, 0]") == 0);
+                          i == 0 ? "[1]" : "[1, 0, 1, 0, 1, 0, 1, 0, 1, 0]") == 0);
 
         uint32_t flip = jit_var_new_op_1(JitOp::Not, mask);
         jit_assert(strcmp(jit_var_str(flip),
-                           i == 0 ? "[0]" : "[0, 1, 0, 1, 0, 1, 0, 1, 0, 1]") == 0);
+                          i == 0 ? "[0]" : "[0, 1, 0, 1, 0, 1, 0, 1, 0, 1]") == 0);
 
         jit_var_dec_ref(flip);
         jit_var_dec_ref(ctr);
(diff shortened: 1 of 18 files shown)


njroussel (approved):
LGTM

tests/basics.cpp
```diff
@@ -86,7 +86,7 @@ TEST_BOTH(03_load_store_mask) {
 }
 
 TEST_BOTH(04_load_store_float) {
-    /// Check usage of floats/doubles (loading, storing, literals)
+    /// Check usage of floats/doubles (loading, storing, constant literals)
```
Speierers, line 89:
```suggestion
    /// Check usage of floats/doubles (loading, storing, literal constants)
```

Speierers (approved):
LGTM

tests/basics.cpp
wjakob: Will fix in a followup commit.

PR #40, opened 2022-11-28
Use consistent ordering of globals (callables & intrinsic declarations) in IR

wjakob:
When Dr.Jit is used in parallel programs, polymorphic classes may be constructed in an arbitrary order. This is problematic since it causes generated callables to shift around in the generated IR representation, which breaks kernel caching. This commit refactors the assembly of globals so that they are always uniquely ordered by their hash key.

This creates the following complication: when a virtual function call is being compiled, we don't yet know the mapping "instance index -> callable index". To determine the ordering of callables we must first generate all of them (including those produced by other virtual function calls) and then sort them.

This commit addresses the issue by introducing a new function `jitc_vcall_upload()` that runs all the way at the end of kernel compilation. At this point, the required information is readily available, and the function can compute and upload the necessary data onto the device.

The generation of globals was refactored to use a central buffer ('globals') and a sorted map ('globals_map') storing offsets into the buffer. Once the kernel has been generated, the sorted entries are copied from 'globals' to 'buffer'. The commit also includes some superficial formatting changes to the generated IR.

### DIFF
--- a/src/cuda_api.cpp
+++ b/src/cuda_api.cpp
@@ -12,8 +12,9 @@
 #include "var.h"
 #include "util.h"
 #include "io.h"
-#include "../resources/kernels.h"
+#include "eval.h"
 #include "cuda_tex.h"
+#include "../resources/kernels.h"
 
 #if defined(_WIN32)
 #  include <windows.h>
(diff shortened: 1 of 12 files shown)


njroussel (approved):
Looks good :rocket:

PR #41, opened 2022-11-29
Make program structure independent of variable indices

wjakob:
Note: this PR builds on PR #40.

When Dr.Jit generates IR output, there is a question of how statements should be ordered. Any ordering that does not violate variable dependencies or basic block boundaries is in principle admissible. Previously, Dr.Jit arranged variables in the order in which they were originally constructed by sorting them according to the monotonically increasing variable index.

This was correct but had three negative implications:

- Variable IDs are 32 bit counters in Dr.Jit, and this is generally desirable to keep the central `Variable` data structure small. However, long-running programs could overflow that 32 bit counter, which meant that the program order no longer made sense. Dr.Jit failed with an error when this happened. At this point, a number of users have encountered this frustrating failure mode.

- Mitsuba, the parent project of Dr.Jit, has the ability to construct scenes in parallel. This leads to an unpredictable ordering of variables and broke kernel caching across runs of the system, hence parallelization had to disabled in practice.

- Variable index changes impeded the effectiveness of the function deduplication optimization.

This commit changes code generation so that IR statements in the generated output are ordered by their scope ID. Within each scope, the order is determined by a DFS traversal.

This is simple and always generates consistent output regardless of variable indices. It no longer matters when the variable index counter overflows.

The same issue will reappear when the scope counter overflows, but 4 billion scopes ought to be ... (ducks).

A few `jit_var_new_scope()` calls had to be inserted in strategic places to ensure that LLVM requirements are satisfied (for example, PHI nodes must always occur at the beginning of a basic block).

### DIFF
--- a/include/drjit-core/state.h
+++ b/include/drjit-core/state.h
@@ -22,9 +22,8 @@ NAMESPACE_BEGIN(detail)
 /**
  * \brief JitState RAII wrapper
  *
- * This class encapsulates several configuration attributes of Dr.Jit (the
- * mask stack, variable name prefixes, the CSE scope, whether program recording
- * is enabled.)
+ * This class encapsulates several configuration attributes of Dr.Jit (the mask
+ * stack, variable name prefixes, whether program recording is enabled.)
  *
  * The <tt>set_*</tt>, <tt>clear_*</tt>, <tt>begin_*</tt>, and <tt>end_*</tt>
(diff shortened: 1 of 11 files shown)


Speierers (approved):
LGTM

Speierers (approved):

njroussel (commented):
LGTM :100:

PR #42, opened 2022-11-29
Improve efficiency of CUDA scatter-reduction

wjakob:
This PR consists of two changes:

The first improves the implementation of the important special case of FP32 scatter-add reduction on CUDA/OptiX. The previous implementation generated up to 32 atomic memory transaction per warp, causing significant contention in some cases. The new implementation first performs an intra-warp reduction. A warp that consistently targets the same location in memory will then only generate a single memory transaction.

The second adjusts the type of PTX IR generated by Dr.Jit. We previously targeted a generic low PTX version and compute capability. The commit changes the logic so that both numbers are bumped according to a table containing known safe combinations.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -2177,13 +2177,10 @@ uint32_t jitc_var_new_scatter(uint32_t target_, uint32_t value, uint32_t index_,
                     "    atom.add.f32 %f19, [%rd3], %f21;\n"
                     "done:\n"
                     "    ret;\n"
-                    "}");
+                    "}\n\n");
 
-                buffer.put("    {\n");
-                buffer.put("        .visible .func reduce_f32(.param .u64 ptr, .param .u32 index, .param .f32 value);\n");
-                buffer.fmt("        call reduce_f32, (%%rd%u, %%r%u, %%f%u);\n",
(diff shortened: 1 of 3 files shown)


src/op.cpp
```diff
@@ -2086,6 +2086,106 @@ uint32_t jitc_var_new_scatter(uint32_t target_, uint32_t value, uint32_t index_,
...
+                    "    vote.sync.any.pred %p8, %p7, %r12;\n"
+                    "    @%p8 bra slow_path;\n"
+                    "    bra.uni maybe_scatter;\n"
+                    "\n"
+                    "fast_path:\n"
+                    "    mov.b32 %r23, %f5;\n"
+                    "    mov.u32 %r24, 2;\n"
+                    "    mov.u32 %r25, 31;\n"
+                    "    mov.u32 %r26, 16;\n"
+                    "    shfl.sync.down.b32 %r27|%p9, %r23, %r26, %r25, %r12;\n"
```
Speierers, line 2148: Really cool to see that `|` operation are allowed on instruction operands directly!

Speierers (approved):
LGTM

njroussel (approved):
LGTM

PR #43, opened 2022-11-30
Fix jitc_var_vcall_reduce for dr.switch()

Speierers:
This PR adapt `jit_var_vcall_reduce` to also work with callable that are not methods of classes registered in the Dr.Jit registry. This is necessary for the implementation of `dr.switch` and `dr.dispatch()`.

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1680,8 +1680,8 @@ struct VCallBucket {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


njroussel (approved):
Can you also update the documentation for `jitc_var_vcall_reduce` in `jit.h` to mention the new behaviour of `bucket_count_out` ?

wjakob:
LGTM.

PR #44, opened 2022-11-30
Bug fix for CUDA assembly & logging improvements for OptiX compilation (minor changes)

njroussel:
The following 2 commits of this PR do:
1) A bug fix related to an issue reported here: https://github.com/mitsuba-renderer/mitsuba3/issues/408. During `jit_assemble_cuda` JIT variables can change addresses because of extra assembly steps required by some variables (a vcall variable in this specific instance). 
2) Improvements to the logging when the OptiX compilation fails. The new message prints the exact failure reported by OptiX rather than only the compilation state.

### DIFF
--- a/src/eval_cuda.cpp
+++ b/src/eval_cuda.cpp
@@ -121,6 +121,7 @@ void jitc_assemble_cuda(ThreadState *ts, ScheduledGroup group,
             if (extra.assemble) {
                 extra.assemble(v, extra);
                 assemble = true;
+                v = jitc_var(index); // The address of 'v' can change
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/eval_cuda.cpp
```diff
@@ -121,6 +121,7 @@ void jitc_assemble_cuda(ThreadState *ts, ScheduledGroup group,
             if (extra.assemble) {
                 extra.assemble(v, extra);
                 assemble = true;
+                v = jitc_var(index); // The address of 'v' can change
```
Speierers, line 124: Nice catch!

Speierers (approved):

PR #45, opened 2022-12-06
Branch/Switch for virtual function calls

njroussel:
This PR adds different strategies to handle virtual function calls.

Currently, in order to do a virtual function call in CUDA, we use indirect function calls to call either an OptiX direct callable or a CUDA function. Given that we know exactly the set of possible targets for any virtual function call, it is not necessary to have the indirection and we can explicitly call the appropriate target function.

This PR adds a new `JitFlag`, called `VCallBranch` which will replace the indirect function call by a series of branches to call the appropriate target function. There are three different branching strategies implemented:

- Linear search: We lineraly compare against all possible target functions to determine the final/appropriate target. This is the default behaviour when enabling `VCallBranch`.
- Binary search: The set of possible target functions is represented as a binary search tree, and the final target is computed by searching through. This strategy can be enabled by using the `VCallBranchBinarySearch` JIT flag.
- Jump table: A jump table from the instance's callable index to all possible targets is used. This strategy can be enabled by using the `VCallBranchJumpTable` JIT flag

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1304,42 +1304,49 @@ enum class JitFlag : uint32_t {
     /// Record virtual function calls instead of splitting them into many small kernel launches
     VCallRecord = 16,
 
+    /**
+     * \brief Use branches instead of direct callables (in OptiX) or indirect
+     * function calls (in CUDA) for virtual function calls.
+     */
+    VCallBranch = 32,
+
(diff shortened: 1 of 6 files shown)


Speierers (commented):
LGTM!

wjakob:
Some years later, the general observation is that compilation via true indirect jumps (and an ABI) seems to be the sweet spot for Dr.Jit that is now consistently implemented across backends.

PR #48, opened 2022-12-08
Minor fixes in LLVM IR

Speierers:


include/drjit-core/state.h
```diff
@@ -90,10 +90,6 @@ template <JitBackend Backend> struct JitState {
         m_prefix_set = false;
     }
 
-    void new_scope() {
```
wjakob, line 93: Aren't we calling this function from the `drjit` parent project?

PR #49, opened 2022-12-12
Fix asynchronous jit_free calls (and more)

Speierers:
This patch fixes a concurrency issue with asynchronous calls to `jitc_free` in `vcall.cpp` and `util.cpp`. 

Here is an example of the potential sequence of events that might lead to a crash:
```
Thread #1: jitc_free(123)
                 -> state.alloc_free.push_back(123);  
Thread #2: jitc_malloc()
                 -> state.alloc_free.find(...); -> 123
                 -> state.alloc_used.push_back(123); -> does nothing since 123 is already in this set
Thread #1: 
                 -> state.alloc_used.erase(123); <- also remove the one that was just added by thread #2
Thread #2: jitc_free(123) -> CRASH since it cannot find 123 in alloc_used as it was erased by thread #1
```

This can be fixed by replacing those two calls to `jitc_free` with `jit_free`.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -316,11 +316,9 @@ void jitc_shutdown(int light) {
                     jitc_var_dec_ref(i);
             }
             jitc_free_flush(ts);
-            if (ts->backend == JitBackend::CUDA) {
+            if (ts->backend == JitBackend::CUDA && ts->stream) {
                 scoped_set_context guard(ts->context);
-                cuda_check(cuEventDestroy(ts->event));
                 cuda_check(cuStreamSynchronize(ts->stream));
-                cuda_check(cuStreamDestroy(ts->stream));
             }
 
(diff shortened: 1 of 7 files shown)


njroussel (approved):
These sort of concurrency bugs are definitely the worse to figure out, great job :tada:

src/init.cpp
```diff
@@ -528,23 +530,20 @@ void jitc_cuda_set_device(int device_id) {
     jitc_log(Info, "jit_cuda_set_device(%i)", device_id);
 
     Device &device = state.devices[device_id];
-    CUcontext new_context = device.context;
 
     /* Disassociate from old context */ {
         scoped_set_context guard(ts->context);
         cuda_check(cuStreamSynchronize(ts->stream));
```
wjakob, line 536: Perhaps add `if (ts->stream)`

PR #50, opened 2022-12-12
Single iteration loop

Speierers:
This patch allows the user to specify that a recorded loop will only perform a single iteration, i which case we avoid executing the condition branch a second time.


njroussel:
@Speierers 

Sorry... I can't figure out, or rather remember, why this was an interesting feature?
(I'm just cleaning up a few old PRs/issues that I unfortunately lost track off...)

Speierers:
Not really sure. I wonder whether this wasn't an attempt at abusing the recording loops to implement `dr.if`/`dr.else` statements.

njroussel:
> abusing the recording loops to implement dr.if/dr.else 

Oh, yes! This was it I think.

PR #51, opened 2022-12-13
Further backend reorganization

wjakob:
This commit refactors the parts of the CUDA and LLVM backends that are related to code generation. The goal is to improve readability, reduce redundancy, and to set the conditions for an eventual transition to a fully abstract IR ("Node")-based representation.

This commit starts with the following changes:

- The `Buffer` class was renamed to `StringBuffer` and moved into separate files `src/stringbuffer.{h,cpp}`.

- The `StringBuffer` class now provides two highly specialized formatting routines for the CUDA and LLVM backends. Their syntax is documented at the top of the `src/eval_{cuda,llvm}.cpp` files. The goal is that these will eventually fully replace the `snprintf`-based function `Buffer::fmt()` to improve performance (`snprintf` is really, really slow) and reduce verbosity (for example, there are specialized escape characters for formatting variables, types, etc.).

- The `Variable` class now supports the backend-agnostic `VarKind::Node` variable kind. Code generation of such abstract IR nodes is handled by node-specific hooks in the corresponding backend (`src/eval_{cuda,llvm}.cpp`). It is expected that these hooks will make heavy use of the new formatting functionality.

- The LLVM backend can now optionally emit IR with opaque pointer types, which is a requirement for supporting newer versions of LLVM. This feature is not yet turned on by default, I plan to add some auto-detection in a later PR to see when it is needed.

- The gather and scatter operations were converted into new abstract IR nodes. The code generation of these is more flexible, and this made it possible to generalize an existing optimization (local SIMD reduction for atomic FP32 scatter-adds) to other operations (multiplication, minimum, maximum, AND, OR) and types.


src/eval_llvm.cpp
```diff
@@ -1,41 +1,103 @@
...
 #include "var.h"
 #include "vcall.h"
 #include "op.h"
 
+#define put(...)                                                               \
+    buffer.put(__VA_ARGS__)
+
+#define fmt(fmt, ...)                                                          \
+    buffer.fmt_llvm(count_args(__VA_ARGS__), fmt, ##__VA_ARGS__)
+
```
Speierers, line 67: Write those macros on a single line?

src/eval_llvm.cpp
```diff
@@ -320,196 +362,297 @@ void jitc_assemble_llvm_func(const char *name, uint32_t inst_id,
...
-        if (vt == (uint32_t) VarType::Float64) {
-            buffer.put(" 0x");
-            buffer.put_uint64_hex(value);
-        } else {
-            buffer.putc(' ');
-            buffer.put_uint64(value);
-        }
+inline bool is_single(const Variable *v) {
+    return (VarType) v->type == VarType::Float32;
+}
```
Speierers, line 418: ditto

src/eval_llvm.cpp
```diff
@@ -320,196 +362,297 @@ void jitc_assemble_llvm_func(const char *name, uint32_t inst_id,
...
+    return (VarType) v->type == VarType::Float64;
+}
+
+inline bool is_uint(const Variable *v) {
+    VarType type = (VarType) v->type;
+    return type == VarType::UInt8 ||
+           type == VarType::UInt16 ||
+           type == VarType::UInt32 ||
+           type == VarType::UInt64;
+}
```
Speierers, line 430: ditto

Speierers (commented):
This new design look really clean! Here are a few comments from my side.

src/strbuf.cpp
```diff
@@ -0,0 +1,659 @@
...
+        if (likely(m_cur + rv < m_end)) {
+            m_cur += rv;
+            return (size_t) rv;
+        } else {
+            expand((size_t) rv);
+        }
+    } while (true);
+}
+
+void StringBuffer::fmt_cuda(size_t nargs, const char *fmt, ...) {
```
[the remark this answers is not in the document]
wjakob: It's a method of `StringBuffer` and those stringification methods are quite large. I would expect that somebody looking at `eval_cuda.cpp` wants to see directly how the main program is created. Another reason for having this function in the same compilation unit is that I want functions like `put_u32_unchecked` to be inlined into the body of `fmt_cuda`.

src/strbuf.cpp
```diff
@@ -0,0 +1,659 @@
...
+                    len += 15;
+                    break;
+
+                case '<':
+                    len += MAXSIZE_U32 + 4;
+                    break;
+
+                case '>':
+                    len += 1;
+                    break;
```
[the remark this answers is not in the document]
wjakob: There are two kinds of switch statements here: one where each case declares local variables, and then the braces are needed. In the other one, there are no local variables, so adding braces seemed like unnecessary syntax noise.

src/eval_cuda.cpp
```diff
@@ -349,90 +369,258 @@ void jitc_assemble_cuda_func(const char *name, uint32_t inst_id,
...
+    Variable *a0 = v->dep[0] ? jitc_var(v->dep[0]) : nullptr,
+             *a1 = v->dep[1] ? jitc_var(v->dep[1]) : nullptr,
+             *a2 = v->dep[2] ? jitc_var(v->dep[2]) : nullptr,
+             *a3 = v->dep[3] ? jitc_var(v->dep[3]) : nullptr;
+
+    switch (v->node) {
+        case NodeType::Add:
+            fmt(is_single(v) ? "    add.ftz.$t $v, $v, $v;\n"
+                             : "    add.$t $v, $v, $v;\n",
+                v, v, a0, a1);
```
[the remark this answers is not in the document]
wjakob: Right -- I was thinking about this as well (and is incidentally how the statement syntax currently works). However, the `fmt_*` methods are now used in many more places than just formatting of the statements -- ray tracing, loops, a future if statement (in fact, it looks like the nodes will end up being the small part). Having this extra indirection would complicate the interface with no benefit to these other use cases.

wjakob:
@Speierers @njroussel This PR now passes all tests (drjit-core, drjit, and mitsuba)! 🎉🎉🎉

src/eval_llvm.cpp
wjakob: One seems too long- this is what clang-format generated, and i broke the short one into two lines for consistency. Or did I misunderstand?

njroussel (approved):
This is very nice, the PTX/LLVM-IR statements are now quite a bit cleaner in the code :partying_face:

PR #52, opened 2022-12-16
Switch to a node-based IR

wjakob:
Building on PR #41, this change refactors the logic in `op.cpp`, `eval_llvm.cpp` and `eval_cuda.cpp` to switch to a node-based IR.

In particular:

- The very large generic function `jit_var_new_op()` was transformed into a set of operation-specific functions like `jit_var_add()`, `jit_var_fma()`. These produce an abstract IR representations, whose codegen-specific bits are now part of the corresponding backends. Splitting the generic function leads to code that is easier to understand and likely easier to compile as well. 

- The somewhat wordy pattern `jit[c]_var_new_x` was changed to the shorter `jit[c]_var_x`, throughout the codebase where `x` is usually an action.

- It is now impossible to create statement-type variables (string-based IR) through the public API. However, internally, there are still a bunch of places that create such variables (CUDA textures, printf, ray tracing, ...). Eventually it will be nice to get rid of these as well, but let's do it in another PR (this one is already way too large)

The PR rolls in a few other minor changes:

- Dr.Jit must know the LLVM version to generate the right set of intrinsics (which change between versions). To do this properly, it now tries harder to detect the version version. When all fails, it uses a fallback that resolves symbols known to only exist in specific LLVM versions, which is good enough for inferring the major version.

- It removes the managed and managed-read-only CUDA memory types that we never used. They are only poorly supported on Windows and have a bunch of performance pitfalls.

- It improves the formatting of some routines (`dr.whos()`, `dr.graphviz()`) that had bit-rotted somewhat.

- Analogous to the LLVM backend, the CUDA and OptiX backends were reorganized into a parts related to dynamic API resolution-specific bits (`{cuda,optix}_api.{h,cpp}`) and the rest (`_core.cpp`)

- I renamed `eval_{cuda,llvm}.cpp` to `{cuda,llvm}_eval.cpp` to follow the naming convention of the other files.

- There was quite a bit of complex code in the memory allocation cache for propagating allocations between different threads that is no longer needed now that each device has a central queue. All of that could be removed, which will hopefully make it easier to support additional backends in the future.

- It turns out that LLVM still had a per-thread queue, which wasn't consistent with how the CUDA backend works. I changed it so that there is also a global queue that different threads submit to. (The minor performance benefits of a per-thread queue are outweighted by the difficulty of getting it to work correctly when mitsuba loads scenes in parallel)


wjakob:
(Updated.)

src/op.cpp
```diff
@@ -1,64 +1,217 @@
...
+    if (size > 0) {
+        // Try simplifying binary expressions with matched arguments
+        if constexpr (Size == 2)
+            simplify |= dep[0] == dep[1];
+
+        for (uint32_t i = 0; i < Size; ++i) {
+            if (unlikely(v[i]->size != size && v[i]->size != 1)) {
+                err = "operands have incompatible sizes";
+                size = (uint32_t) -1;
+                goto fail;
```
njroussel, line 160: This might actually be the first time I see a `goto` in real code :smile:

njroussel (approved):
A few typos, but overall this is very nice :100: 
Especially, `src/op.cpp` has a really clean structure now!

src/op.cpp
wjakob: I think one is entitled to use one every now and then :D (very handy for a centralized failure/cleanup part of a function). It's very often used in the CPython codebase, e.g.: https://github.com/python/cpython/blob/main/Objects/longobject.c#L1038

wjakob:
I did an interactive rebase to squash some sub-commits and integrated your feedback. I added one more minor change (https://github.com/mitsuba-renderer/drjit-core/pull/52/commits/a3c7e5bef28e06dfe79e42dba3917645992241ab) to change the `printf_async` operation from textual IR into abstract IR.

src/op.cpp
njroussel: This rings a bell, for a completely different reason

There was a CppCon talk about some `goto` trickery that could boost performance significantly. I remember the speaker mentionning that CPython made use of this.

Issue #53, opened 2023-01-31
Undefined behavior in StringBuffer


dvicini:
Hi @wjakob, hi @njroussel

We have been compiling drjit with various Clang sanitizers enabled on our end. We use these sanitizers pretty much by default to guard against some subtle issues due to undefined behavior and invalid memory accesses. 

More concretely, I am using Clang 14 and the following cmake command:
```
cmake -E env CXXFLAGS="-fsanitize=address,undefined" cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DDRJIT_ENABLE_TESTS=1
```

There are a few places in the code where some undefined behavior occurs. I think an important one to consider is the `StringBuffer` class. In this class, a few times there is some pointer arithmetic to determine validity of write operations, e.g. lines like
```
if (likely(m_cur + rv < m_end)) {
...
```
https://github.com/mitsuba-renderer/drjit-core/blob/master/src/strbuf.cpp#L198

This is generally fine, but problematic for empty StringBuffers, where `m_cur` is a nullptr. The issue is that pointer arithmetic using a null pointer is actually undefined behavior. Using the sanitizers, I get a runtime error when running `test_basics` on that line.

I was wondering if it would make sense to just change the constructor of the StringBuffer class to ensure an internal capacity of at least 1? This would prevent m_cur ever being a null pointer and would not require any further changes in the file or anywhere else. I also assume that there are no important places where we rely on using completely empty string buffers?

Best,
Delio

dvicini:
Closing this as we can discuss in #55 if needed

PR #54, opened 2023-01-31
Pad cache file contents to prevent misaligned memory accesses

dvicini:
Hi, 

When running drjit-core with sanitizers enabled (see also issue #53 for more context), we spotted a misaligned memory read/write in the cache file loading and writing routine. 

In particular, the part of the code handling the relocation table performs writes that aren't necessarily aligned, depending on the content that was written previously. This raises an error when checking the code with the sanitizer and theoretically could be a practical problem down the line.

This PR contains a fix that simply pads the content of the file with zeros to ensure alignment. I am happy to iterate on this if you have a better solution in mind.

Best,
Delio

### DIFF
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -28,7 +28,7 @@
 #endif
 
 /// Version number for cache files
-#define DRJIT_CACHE_VERSION 4
+#define DRJIT_CACHE_VERSION 5
 
 // Uncomment to write out training data for creating a compression dictionary
 // #define DRJIT_CACHE_TRAIN 1
@@ -61,6 +61,15 @@ void jitc_lz4_init() {
         jitc_fail("jit_init_lz4(): decompression of dictionary failed!");
 
     jitc_lz4_dict_ready = true;
+}
+
+/* Computes padding to align cache file content to a multiple of sizeof(void*). 
+This prevents undefiend behavior due to misaligned memory reads/writes. */
+static uint32_t compute_padding(const CacheFileHeader &header) {
+    uint32_t padding_size = (header.source_size + header.kernel_size) % sizeof(void *);
+    if (padding_size)
+        padding_size = sizeof(void *) - static_cast<int>(padding_size);
+    return padding_size;
 }
 
 bool jitc_kernel_load(const char *source, uint32_t source_size,
@@ -134,6 +143,7 @@ bool jitc_kernel_load(const char *source, uint32_t source_size,
     char *compressed = nullptr, *uncompressed = nullptr;
 
     CacheFileHeader header;
+    uint32_t padding_size;
     bool success = true;
     std::vector<void *> func;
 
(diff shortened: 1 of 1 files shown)


wjakob:
Hi Delio,

that's quite interesting, perhaps we should do that as well (use of sanitizers). First, a question: is there a more, or are those two issues reported sufficient to silence the sanitizers. 

Regarding this specific PR: padding the relocations is fine,  but let's do it without adding a (redundant) field to the header. The actual offset is easy to infer, both on the reader and the writer side. It would be great if you could add a comment, e.g. "Pad to a multiple of void* to avoid platform-specific undefined behavior when reading/writing pointers". The file format version will need a bump, or there will be cache conflicts.

wjakob:
More questions :-) What are the GCC/Clang flags that you add to the compilation? Do you know the typical impact on execution time?

dvicini:
Hi Wenzel, 

The two issues I reported are the more serious ones in terms of correctness, but there are a few other ones. Those are a all pretty minor as far as I can tell. If you are interested in having those fixed, I am very happy to make a PR. Either way we have fixed a few of them on our end, so the more of those we can port back in the open source version, the better (this will make it easier for us to stay in sync with future drjit/drjit-core updates)

You should be able to see them by running
```
cmake -E env CXXFLAGS="-fsanitize=address,undefined" cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DDRJIT_ENABLE_TESTS=1
```
and then just running the tests as usual. I compile with Clang, but I think even with GCC this does something useful. 

I don't have a feel for the impact on execution time. It seems to run fast enough on my machine to run the drjit-core tests, but I haven't tried in a more complex setup (i.e. Mitsuba). So far I've been positively surprised by how well these flags allow to catch a range of undefined behaviors and also some of the smaller memory leaks that drjit-core has (actually some of those leaks are a bit trickier to fix than simple undefined behavior)


I will fix this PR as you suggested, indeed the offset can easily be computed without the extra field I think.

dvicini:
I just pushed an improved version of the fix. I added a small helper function to compute the padding the on the fly and also bumped the DRJIT_CACHE_VERSION.

If this looks ok to you we can merge this and I can make another PR addressing other minor undefined behaviors

src/io.cpp
```diff
@@ -63,6 +63,15 @@ void jitc_lz4_init() {
     jitc_lz4_dict_ready = true;
 }
 
+/* Computes padding to align cache file content to a multiple of sizeof(void*). 
+This prevents undefiend behavior due to misaligned memory reads/writes. */
+uint32_t compute_padding(const CacheFileHeader &header) {
```
njroussel, line 68:
```suggestion
static uint32_t compute_padding(const CacheFileHeader &header) {
```

njroussel (commented):
One tiny change, then I think we're good to merge this  :+1:

wjakob:
Hi @dvicini  --- I'm seeing a concerning-looking warning that appear to result from this change on MSVC:

C:\BuildAgent\work\d2336b22797e7f25\src\io.cpp(215): warning C4701: potentially uninitialized local variable 'padding_size' used [C:\BuildAgent\work\d2336b22797e7f25\build\drjit-core.vcxproj]

Could I ask you to take a look to see if this is benign? (perhaps it just needs a small change by zero-initializing something so that the warning goes away?)

dvicini:
I took a look and it is benign. What happens is that there might be an exception in the block where pad size is initialized, setting success = false. In that case, the code after is either way skipped (including the line 215). I pushed a commit that adds a zero initialization, which silences the warning.

PR #55, opened 2023-02-01
Various fixes of undefined behavior

dvicini:
This PR fixes a few smaller undefined behaviors (see also the discussion in #54).

The fixes in this PR so far are very minor: 
- it ensures that StringBuffers always have a capacity of >= 1. This fixes undefined behavior related to pointer arithmetic on null pointers (addresses #53)
- it fixes an undefined behavior due to invoking memcpy with a nullptr by first checking if >0 bytes should be copied. Calling memcpy(..., nullptr, 0) is invalid, even though logically this should be fine. 
- it fixes a small memory leak of an LLVM error string

There are a few more issues that we we could address also in this PR:
- In test_basics, we are testing the result of CTZ and SHL operations using negative numbers. Left-shifting negative numbers is undefined behavior. Maybe these ops should be skipped for negative arguments in the tests? Not sure whats a good way to do that without overcomplicating the test code though. 
- There are a few more memory leaks. The main one appears to be the `void **reloc` field inside the `Kernel` struct. What would be a good way to fix this? I wasn't sure what would make sense in the context of the code base (i.e. which C++ constructs we use here and which we avoid, e.g. could this field just be an `std::vector` instead?).

### DIFF
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -443,6 +443,8 @@ bool jitc_kernel_write(const char *source, uint32_t source_size,
 
 void jitc_kernel_free(int device_id, const Kernel &kernel) {
     if (device_id == -1) {
+        if (kernel.llvm.n_reloc)
+            free(kernel.llvm.reloc);
 #if !defined(_WIN32)
         if (munmap((void *) kernel.data, kernel.size) == -1)
             jitc_fail("jit_kernel_free(): munmap() failed!");
--- a/src/llvm_core.cpp
+++ b/src/llvm_core.cpp
@@ -327,7 +327,8 @@ void jitc_llvm_compile(Kernel &kernel) {
     if (unlikely(error))
         jitc_fail("jit_llvm_compile(): parsing failed. Please see the LLVM "
                   "IR and error message below:\n\n%s\n\n%s", buffer.get(), error);
-
+    LLVMDisposeMessage(error);
+    
 #if !defined(NDEBUG)
     bool status = LLVMVerifyModule(llvm_module, LLVMReturnStatusAction, &error);
     if (unlikely(status))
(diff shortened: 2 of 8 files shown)


njroussel:
I'm a bit suprised that arithmetic with `nullptr` is UB. Anyway, I'd  prefer if the constructor threw an error or if we explicitly handled the 0-capacity/`nullptr`-case wherever necessary (there's a call to `expand()` in most methods which should make it safe for use after). Setting the size to 1 is  _hacky_.

The other two fixes are fine with me.

Regarding the CTZ/SHL operations, I don't see how this would over-complicate the code. Isn't it just a matter of adding something like `if(isSigned && op == JitOp::Shl && values[i] < 0) continue;` at two spots?

As for the memory leaks, it would be nice to not include a huge STL header for something like this. For the `Kernel` struct specifically, the missing deallocations should all take place in `jitc_kernel_free()`. Or have you identified some other code path?

dvicini:
1. Yes I agree that the fix isn't so elegant. Maybe indeed throwing an error and fixing all the users is a better idea. I don't think the buffer can ever shrink, so as long as it's initialized to capacity >= 1 this should be safe
2. CTZ/SHL: Okay, maybe I missed something there. Somehow it looked more complicated at first.
3. I think that could make sense. I first thought we didn't have a way of telling whether the kernel has the reloc field allocated, but I guess we can infer this from the device_id in the free function. I can take a look when i find some time

dvicini:
@njroussel 

Accidentally closed this PR yesterday, now re-opened with updated fixes. I fixed the "reloc" memory leak and adjusted the stringbuffer fix.

There are two more issues I haven't figured out a good fix for yet:
- the CTZ function internally can do a left shift of a negative number even if the argument was originally a positive, signed integer. Not sure what's the best way to fix this. It's not just a matter of not testing negative inputs. 
- there is a memory leak in the loop.name variable that appears when running test_loops. I think it only happens with LoopOptimize = true, but I couldn't localize it further. Probably easy to fix, but I don't have the full picture of the loop optimization

njroussel (approved):

PR #56, opened 2023-02-14
Implement Kahan-style scatter-accumulation for LLVM

njroussel:
This PR implements the missing logic for `VarKind::ScatterKahan` in LLVM mode.

(Run for CI)


wjakob:
Thanks @njroussel, this looks great. Nice catch with the variable name typo.

PR #59, opened 2023-04-07
Add emulation for CPU without FMA instruction set

lnuic:
# Description 

FMA instruction set was introduced in 2012 (AMD Piledriver, Intel Haswell),  but architectures before that already had AVX and SSE4.2 instruction sets.  Dr. Jit doesn't currently verify whether a CPU supports the FMA instruction set. As a result, LLVM generates a global offset table when FMA is not available, which leads to a critical compiler failure in Dr. Jit.

To address this issue, emulation of the FMA instruction was implemented using the existing fmul and fadd instructions. This will ensure that the code runs smoothly on CPUs that do not have native FMA support, preventing any potential failures caused by the absence of the instruction set.

This PR should also fix: [mitsuba-renderer/drjit/#46](https://github.com/mitsuba-renderer/drjit/issues/46)

## Code to reproduce:

```
import drjit as dr
dr.set_log_level(dr.LogLevel.Trace)
a = dr.llvm.Array3f([1.], [2.], [3.])
b = dr.fma(a, a, a)
print(b)
```

## Error message:
```
Critical Dr.Jit compiler failure: jit_llvm_compile(): a global offset table was generated by LLVM, which typically means that a compiler intrinsic was not supported by the target architecture. DrJit cannot handle this case and will terminate the application now. For reference, the following kernel code was responsible for this problem:

define void @drjit_4300624cc1a9b63fa571334796cbc01c(i64 %start, i64 %end, i8** noalias %params) #0 {
entry:
    br label %body

body:
    %index = phi i64 [ %index_next, %suffix ], [ %start, %entry ]
    %f1_p1 = getelementptr inbounds i8*, i8** %params, i32 3
    %f1_p2 = load i8*, i8** %f1_p1, align 8, !alias.scope !2
    %f1_p3 = bitcast i8* %f1_p2 to float*
    %f1_0 = load float, float* %f1_p3, align 4, !alias.scope !2
    %f1_1 = insertelement <8 x float> undef, float %f1_0, i32 0
    %f1 = shufflevector <8 x float> %f1_1, <8 x float> undef, <8 x i32> zeroinitializer
    %f2_p1 = getelementptr inbounds i8*, i8** %params, i32 4
    %f2_p2 = load i8*, i8** %f2_p1, align 8, !alias.scope !2
    %f2_p3 = bitcast i8* %f2_p2 to float*
    %f2_p4 = getelementptr inbounds float, float* %f2_p3, i64 %index
[...]

PR #60, opened 2023-05-02
Disable LLVM backend when the FMA3 instruction set is unavailable

njroussel:
This PR is a follow-up to #59.

When the FMA3 instruction set is not supported on the target CPU, the JIT will now gracefully shutdown with a warning message.

Correctly emulating the instruction is difficult and bumping the CPU architecture requirement is too-strict as some users might only use the CUDA backend.

(Fixes https://github.com/mitsuba-renderer/drjit/issues/46).

### DIFF
--- a/src/llvm_core.cpp
+++ b/src/llvm_core.cpp
@@ -103,6 +103,14 @@ bool jitc_llvm_init() {
         }
     }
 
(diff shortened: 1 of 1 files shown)


njroussel:
@lnuic ping for visibility

You can keep using your workaround/emulation locally - it's simply not something we want to support/maintain globally :slightly_frowning_face:

PR #66, opened 2023-08-30
Add OptiX debug/validation level as CMake options

rtabbara:
Add support to optionally enable debug and validation flags in OptiX


njroussel (approved):
Nice! :+1:

PR #68, opened 2023-09-12
Fix loading Bool during VCall

mcrescas:
This PR fixes a specific issue when loading a Bool type during a VCall using LLVM.
Specifically, a type transformation was missing leading to invalid IR code.


njroussel:
Thanks :+1: 

I've pushed a slightly different fix with a test (https://github.com/mitsuba-renderer/drjit-core/commit/8ecaaf28aa62cc5b2724385f11b039c4b430e74e).

PR #69, opened 2023-10-02
Fix a typo in the name of `nvidia-smi` utility.

c86-moe:
This commit fixed a typo in the name of the `nvidia-smi` utility, a term which users may elect to copy and run in their terminal during troubleshooting.

There are no other language or functionality changes in this PR.


njroussel:
Hi @c86-moe 

Thanks!

PR #70, opened 2023-10-25
Adding fp16 support

rtabbara:


src/cuda_eval.cpp
```diff
@@ -194,7 +197,7 @@ void jitc_cuda_assemble(ThreadState *ts, ScheduledGroup group,
                 fmt("    mad.wide.u32 %rd0, %r0, $a, %rd0;\n", v);
 
             if (vt != VarType::Bool) {
-                fmt("    $s$t $v, [%rd0];\n",
```
wjakob, line 197: Would the existing `$b` not address the problem in a more general way here?

include/drjit-core/half.h
```diff
@@ -0,0 +1,119 @@
+/*
+    drjit-core/half.h -- minimal half precision number type
+
+    Copyright (c) 2021 Wenzel Jakob <wenzel.jakob@epfl.ch>
```
wjakob, line 4: Hm, it's weird to list me as copyright author given that you wrote this code. Why don't you remove it for now or put your name? Or we can change the copyright note to list the lab in a later commit.

src/cuda_eval.cpp
```diff
@@ -214,7 +217,7 @@ void jitc_cuda_assemble(ThreadState *ts, ScheduledGroup group,
                 params_type, params_base, v, v);
 
             if (vt != VarType::Bool) {
-                fmt("    st.global.cs.$t [%rd0], $v;\n", v, v);
+                fmt("    st.global.cs.$T [%rd0], $v;\n", v, v);
```
wjakob, line 220: ditto

src/cuda_eval.cpp
```diff
@@ -336,7 +339,7 @@ void jitc_cuda_assemble_func(const char *name, uint32_t inst_id,
 
         if (v->vcall_iface) {
             if (vt != VarType::Bool) {
-                fmt("    ld.param.$t $v, [params+$o];\n", v, v, v);
+                fmt("    ld.param.$T $v, [params+$o];\n", v, v, v);
```
wjakob, line 342: ditto

src/cuda_eval.cpp
```diff
@@ -359,7 +362,7 @@ void jitc_cuda_assemble_func(const char *name, uint32_t inst_id,
                     "is happening now). This is not allowed.", sv.index);
 
             if (vt != VarType::Bool)
-                fmt("    ld.global.$t $v, [data+$u];\n",
+                fmt("    ld.global.$T $v, [data+$u];\n",
```
wjakob, line 365: ditto

src/cuda_eval.cpp
```diff
@@ -383,7 +386,7 @@ void jitc_cuda_assemble_func(const char *name, uint32_t inst_id,
         uint32_t vti = v->type;
 
         if ((VarType) vti != VarType::Bool) {
-            fmt("    st.param.$t [result+$u], $v;\n", v, offset, v);
+            fmt("    st.param.$T [result+$u], $v;\n", v, offset, v);
```
wjakob, line 389: ditto

src/cuda_eval.cpp
```diff
@@ -430,8 +433,17 @@ static void jitc_cuda_render_var(uint32_t index, const Variable *v) {
             break;
 
         case VarKind::Sqrt:
-            fmt(jitc_is_single(v) ? "    sqrt.approx.ftz.$t $v, $v;\n"
-                                  : "    sqrt.rn.$t $v, $v;\n", v, v, a0);
+            if (jitc_is_half(v))
+                fmt("    cvt.f32.f16 %f0, $v;\n"
```
wjakob, line 437: Given that this pattern repeats in a few places, how about we have a special per-instruction flag stating whether such a conversion is needed? Then those two lines could be added before/after the switch statement. And we can temporarily overwrite the `v->reg_index` so that `$v` reaches the right operand.

 We'll need to reuse this approach if we ever add 8-bit arithmetic, etc.

wjakob (commented):
Hi Rami,

this already looks fantastic and is nice and short. A few first quick comments from me:

tests/loop.cpp
```diff
@@ -11,13 +11,13 @@ TEST_BOTH(01_record_loop) {
         for (uint32_t j = 0; j < 2; ++j) {
             UInt32 x = arange<UInt32>(10);
             Float y = zero<Float>(1);
-            Float z = 1;
+            Float z = Float(1);
```
wjakob, line 14: I am curious: why does this not work? Having an implicit conversion from an integer to a float is probably a good idea.

src/op.cpp
```diff
@@ -291,10 +295,10 @@ uint32_t jitc_var_shift(const VarInfo &info, uint32_t index, uint64_t amount) {
 
 // --------------------------------------------------------------------------
 
-template <typename T, enable_if_t<!std::is_signed_v<T>> = 0>
+template <typename T, enable_if_t<!std::is_signed<T>::value> = 0>
```
wjakob, line 298: I am curious: why was this change needed? The two should be the same, and the former version was shorter.

include/drjit-core/half.h
```diff
@@ -0,0 +1,117 @@
...
+template<> struct is_signed<drjit::dr_half> : true_type { };
+NAMESPACE_END(std)
+
+
+NAMESPACE_BEGIN(drjit)
+struct dr_half {
+
+    uint16_t value;
+
+    #define DRJIT_IF_SCALAR template <typename Value, enable_if_t<std::is_arithmetic<Value>::value> = 0>
```
njroussel, line 41:
```suggestion
    #define DRJIT_IF_SCALAR template <typename Value, enable_if_t<!std::is_arithmetic_v<Value>>
```

include/drjit-core/half.h
```diff
@@ -0,0 +1,117 @@
...
+
+    dr_half() = default;
+
+    DRJIT_IF_SCALAR dr_half(Value val) : value(float32_to_float16(float(val))) { }
```
njroussel, line 45: Using `float()` as a constructor rather than a C-style cast is a bit weird here. No?

src/cuda_eval.cpp
```diff
@@ -1271,16 +1308,16 @@ void jitc_var_vcall_assemble_cuda(VCall *vcall, uint32_t vcall_reg,
         const Variable *v2 = &it->second;
         uint32_t size = type_size[v2->type];
 
-        const char *tname = type_name_ptx[v2->type],
-                   *prefix = type_prefix[v2->type];
+        const char *tname               = type_name_ptx_bin[v2->type],
+                   *prefix              = type_prefix[v2->type];
```
njroussel, line 1312: No need to align these

src/llvm_eval.cpp
```diff
@@ -733,10 +733,22 @@ static void jitc_llvm_render_var(uint32_t index, Variable *v) {
                                     : "    $v = fptosi $V to $T\n",
                     v, a0, v);
             } else if (jitc_is_float(v) && jitc_is_float(a0)) {
-                fmt(type_size[v->type] > type_size[a0->type]
-                        ? "    $v = fpext $V to $T\n"
-                        : "    $v = fptrunc $V to $T\n",
-                    v, a0, v);
+                // On x86, direct double-half casting relies on external builtin function call
+                // unless AVX512_FP16 instructions are supported so split casting into two-steps
+                // i.e. double<->float<->half
```
njroussel, line 738: :+1:

src/op.cpp
```diff
@@ -291,10 +295,10 @@ uint32_t jitc_var_shift(const VarInfo &info, uint32_t index, uint64_t amount) {
...
 T eval_neg(T v) { return T(-(std::make_signed_t<T>) v); }
 
-template <typename T, enable_if_t<std::is_signed_v<T>> = 0>
+template <typename T, enable_if_t<std::is_signed<T>::value> = 0>
```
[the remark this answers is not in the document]
njroussel: Nevermind, there are quite a few more on the following lines

src/op.cpp
```diff
@@ -341,7 +345,7 @@ uint32_t jitc_var_not(uint32_t a0) {
 // --------------------------------------------------------------------------
 
 template <typename T, enable_if_t<std::is_floating_point_v<T>> = 0>
-T eval_sqrt(T value) { return std::sqrt(value); }
+T eval_sqrt(T value) { using std::sqrt, drjit::dr_half; return ::sqrt(value); }
```
njroussel, line 348: Very clean!

src/var.cpp
```diff
@@ -410,6 +410,7 @@ void jitc_value_print(const Variable *v, bool graphviz = false) {
         break;
 
     switch ((VarType) v->type) {
+        case VarType::Float16: JIT_LITERAL_PRINT(uint16_t, unsigned, "%u");
```
njroussel, line 413: `jitc_value_print` is meant for human-readable outputs. I think it would make sense if printed the floating point value rather than its `uint16_t` representation.

njroussel: I see now that you're doing this in `jitc_var_str()`. I think we would something similar here.

include/drjit-core/half.h
njroussel: We should also maybe explicitly instantiate this constructor for all arithmetic types. Otherwise we won't get any implicit constructor calls like `dr_half my_var = 3u;` in other libraries.

njroussel (commented):
Looking good! I added a few suggestions for improvements :+1:

wjakob:
How about we rename the half type from `dr_half` to just `half`?

wjakob:
A stupid question: out of all the various arithmetic operations, why is sqrt treated specially?

src/cuda_eval.cpp
rtabbara: Yes, you're right..I've fixed that up.

src/cuda_eval.cpp
rtabbara: I was trying to be a bit too clever by adding a special type character just to handle the fact PTX instructions for a lot of the data move/load/store operations can't be specified by a `fp16`  type and instead need `b16` (keeping all the other types the same). But I agree, just having `$b` for everything doesn't really hurt readability and reduces bloat so I've made that switch.

src/cuda_eval.cpp
rtabbara: Makes sense, that change has been added!

tests/loop.cpp
rtabbara: This has been fixed

src/op.cpp
rtabbara: Specializing `std::is_signed`, `std::is_arithmetic` etc. is officially considered undefined behaviour, which we were using for the `half` type. For GCC `is_signed_v` worked fine, but then I was having issues compiling with Clang which is why I made this change. But then I realised MSVC was also broken and in the end I thought it was safest to define our own `drjit::is_signed` etc. traits to avoid relying on undefined behaviour. So hope that's ok.

include/drjit-core/half.h
```diff
@@ -0,0 +1,117 @@
...
+        #elif defined(__aarch64__)
+            return (float)memcpy_cast<__fp16>(value);
+        #else
+            jit_fail("Unsupported architecture");
+        #endif
+    }
+
+    static dr_half sqrt(dr_half h) { 
+        #if defined(__aarch64__)
+            return (float)vsqrth_f16(memcpy_cast<__fp16>(h.value));
```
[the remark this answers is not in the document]
rtabbara: That's a good point. Fixed!

src/op.cpp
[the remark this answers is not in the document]
rtabbara: As mentioned before, this was my hacky way to get around undefined behaviour of specialising these type traits

include/drjit-core/half.h
rtabbara: Fixed!

include/drjit-core/half.h
```diff
@@ -0,0 +1,114 @@
...
+#include <drjit-core/jit.h>
+#include <drjit-core/traits.h>
+
+#include <cmath>
+#include <cstring>
+
+#if defined __aarch64__
+#include <arm_fp16.h>
+#elif defined(__F16C__)
+#include <emmintrin.h>
```
wjakob, line 19: I think you can remove `<emmintrin.h>` here. AFAIK `<immintrin.h>` includes everything.

src/cuda_eval.cpp
```diff
@@ -408,6 +408,26 @@ static void jitc_cuda_render_var(uint32_t index, const Variable *v) {
              *a2 = v->dep[2] ? jitc_var(v->dep[2]) : nullptr,
              *a3 = v->dep[3] ? jitc_var(v->dep[3]) : nullptr;
 
+    bool f32_upcast = jitc_is_half(v) && var_kind_fp16_unsupported_cuda[v->kind];
+
+    if (f32_upcast) {
+        Variable* b = const_cast<Variable*>(v);
+        b->type = (uint32_t)VarType::Float32;
+        for (size_t i = 0; i < 4; ++i) {
+            Variable* dep = b->dep[i] ? jitc_var(b->dep[i]) : nullptr;
```
wjakob, line 417: (Despite the comment below). Would it not be easier to understand to merge these two loops? You can add a check to see if the variable has already been set to `Float32`. I think the current implementation may generate the `cvt.f32.f16` instruction multiple times for operations that receive the same operand twice.

src/cuda_eval.cpp
```diff
@@ -802,6 +842,8 @@ static void jitc_cuda_render_scatter(const Variable *v,
     const char *op = reduce_op_name[v->literal];
     const ThreadState *ts = thread_state_cuda;
 
+    if (jitc_is_half(value) && v->literal != 0)
```
wjakob, line 845: This is not a nice way to catch this special case because it will crash the Python session. Can we already do this in `op.cpp` where the associated scatter variable would be created?

src/cuda_eval.cpp
```diff
@@ -920,15 +964,16 @@ static void jitc_cuda_render_scatter_kahan(const Variable *v, uint32_t v_index)
...
+    const char* red_atom_suffix = jitc_is_half(value) ? ".noftz" : "";
 
     fmt("    {\n"
         "        .reg.$t %before, %after, %value, %case_1, %case_2;\n"
         "        .reg.$t %abs_before, %abs_value, %result;\n"
         "        .reg.pred %cond;\n"
         "\n"
-        "        mov.$t %value, $v;\n"
-        "        atom.global.add.$t %before, [%rd2], %value;\n"
+        "        mov.$b %value, $v;\n"
```
wjakob, line 975: If float16 atomics aren't properly supported, then I don't understand why there are changes here. I would prefer to keep the patch small in this case, unless I am misunderstanding something?

src/var.cpp
```diff
@@ -172,6 +172,94 @@ const char *var_kind_name[(int) VarKind::Count] {
     "extract"
 };
 
+const bool var_kind_fp16_unsupported_cuda[(int) VarKind::Count] {
```
wjakob, line 175: would it not be easier to understand to encode the inverse? (i.e., `fp16_supported`)

wjakob (commented):
Hi Rami,

Thanks for making those changes — here are a few more comments from me.

-Wenzel

src/op.cpp
wjakob: Making our own specialized version of these classes seems fine and safest.

PR #71, opened 2023-11-27
Additional fp16 changes to support CUDA-textures

rtabbara:
* Also adding custom scatter reduce (add) for fp16 types


src/cuda_eval.cpp
```diff
@@ -763,22 +763,32 @@ static void jitc_cuda_render_var(uint32_t index, Variable *v) {
             break;
 
         case VarKind::TexLookup:
-            fmt("    .reg.f32 $v_out_<4>;\n", v);
+            fmt("    .reg.$t $v_out_<4>;\n", v, v);
             if (a3)
-                fmt("    tex.3d.v4.f32.f32 {$v_out_0, $v_out_1, $v_out_2, $v_out_3}, [$v, {$v, $v, $v, $v}];\n",
-                    v, v, v, v, a0, a1, a2, a3, a3);
+                fmt("    tex.3d.v4.$t.f32 {$v_out_0, $v_out_1, $v_out_2, $v_out_3}, [$v, {$v, $v, $v, $v}];\n",
```
wjakob, line 768: Out of curiosity: can CUDA also perform texture lookups that produce a f16 output?

src/cuda_eval.cpp
```diff
@@ -964,6 +975,23 @@ static void jitc_cuda_render_scatter(const Variable *v,
...
         );
+    } else if (v->literal && is_half) {
+        // Encountered OptiX link errors attempting to use red.global.add.noftz.f16
+        // so use f16x2 instead
+        fmt("    {\n"
+            "        .reg .f16x2 %packed;\n"
+            "        .reg .b64 %align, %offset;\n"
+            "        .reg .b32 %offset_32;\n"
+            "        .reg .f16 %initial;\n"
+            "        mov.b16 %initial, 0;\n"
```
wjakob, line 986: Great! This looks excellent.

wjakob (approved):
Looks all good to me!

src/cuda_eval.cpp
rtabbara: I dad quick look at the CUDA documentation and found:

> 16-bit floating-point components are promoted to 32 bit float during texture fetching before any filtering is performed.

This also aligns with the fact that the PTX instruction fetch `tld4` doesn't support `f16` destinations. So my guess is that internally a lookup will be f32 and then converted to f16? Still, it's nice that the PTX lookup instruction abstracts away those steps.

PR #75, opened 2024-01-08
Add jit_cuda_sync_stream operation

rtabbara:


src/api.cpp
```diff
@@ -287,6 +287,11 @@ void *jit_cuda_lookup(const char *name) {
     return jitc_cuda_lookup(name);
 }
 
+void jit_cuda_sync_stream(uint64_t stream) {
+    lock_guard guard(state.lock);
```
wjakob, line 291: No synchronization is needed for this operation, since you are only accessing the thread local state, but not the `state` variable. You can leave a comment saying so and remove the `lock_guard`.

src/cuda_api.cpp
```diff
@@ -197,4 +197,12 @@ void *jitc_cuda_lookup(const char *name) {
         jitc_raise("jit_cuda_lookup(): function \"%s\" not found!", name);
     return ptr;
 }
+
+void jitc_cuda_sync_stream(uint64_t stream) {
+    ThreadState* ts = thread_state(JitBackend::CUDA);
+    CUevent sync_event = ts->sync_stream_event;
+    cuda_check(cuEventRecord(sync_event, ts->stream));
```
wjakob, line 204: Why is the cast to `CUstream` needed in one place, but not the other?

wjakob:
LGTM!

njroussel (approved):
Looks gooooodd :+1:

PR #80, opened 2024-02-14
Fix race condition in `jit_init`

merlinND:
A race condition was introduced in 12a2b8b: `jit_init` calls `jitc_cuda_compile`, which now locally unlocks the state lock. This means that operations such as `jitc_init_thread_state` could start executing before `jit_init` would complete.

The race condition was easily reproducible on my machine by deleting the cache directories. The next run would almost always trigger:

```
jit_init_thread_state(): the CUDA backend is inactive because it has not been initialized via jit_init(), or because the CUDA driver library ("libcuda.so") could not be found! Set the DRJIT_LIBCUDA_PATH environment variable to specify its path.
```

As a simple fix, I introduced a `release_state_lock` argument, which is only set to `false` in `jit_init`. Hopefully that's the only case where we really need to keep the lock?

Also fixed some indentation (mixed tabs and spaces).

---

PS: technically I encountered this after backporting 12a2b8b to `master`, but if I understand correctly, the `nanobind` branch is affected as well.

### DIFF
--- a/src/cuda.h
+++ b/src/cuda.h
@@ -39,7 +39,8 @@ extern void *jitc_cuda_lookup(const char *name);
 struct Kernel;
 
 /// Compile an PTX string. Returns the resulting module and a cache hit true/false flag
-extern std::pair<CUmodule, bool> jitc_cuda_compile(const char *str);
+extern std::pair<CUmodule, bool>
+jitc_cuda_compile(const char *str, bool release_state_lock = true);
 
 /// Assert that a CUDA operation is correctly issued
 #define cuda_check(err) cuda_check_impl(err, __FILE__, __LINE__)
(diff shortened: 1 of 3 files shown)


merlinND:
The CI failures don't look related to this PR:

```
  /home/ci-agent/TeamCity/temp/agentTmp/custom_script17281547226288108668: 2: tests/test_loop: not found
```

wjakob:
Great, thank you Merlin. You are right that the CI is currently broken.

wjakob:
How did you manage to find this by the way? Are you running some kind of thread sanitizer?

merlinND:
In my case it occurred very reliably without trying, as soon as I cleared the kernel cache directories. Given the error message (`the CUDA backend is inactive because it has not been initialized via jit_init()`) I thought it was due to some driver installation / kernel module error at first, but by checking the logic of `jit_init` it became clear.

wjakob:
Got it, thank you for reporting and fixing the issue. 👍

PR #81, opened 2024-02-23
Formatter for pointer to const types

njroussel:
Pointer to const types did not have a specialization for an appropriate `formatter`. 

:warning: This requires an include of `type_traits` in the `nanostl.h` file. (Technically, we were already using `std::is_pointer_v` which is defined in `type_traits`, so I'm guessing `utility` already partially covers `type_traits`.)


wjakob:
LGTM!

PR #83, opened 2024-03-07
ThreadState Interface

DoeringChristian:
Made the `ThreadState` struct an interface with `CUDAThreadState` and `LLVMThreadState` implementing it.
Added virtual member functions for all functions in `util.cpp`.
Implemented those functions in the `CUDA` and `LLVM` variants.
The original functions still exist and directly call the ones in `ThreadState` for compatibility.


CMakeLists.txt
```diff
@@ -134,6 +135,7 @@ add_library(
   src/llvm_eval.cpp
   src/llvm_scatter.h
   src/llvm_scatter.cpp
+  src/llvm_ts.cpp
```
wjakob, line 138: Can you add the header file here as well?

wjakob: ditto for the CUDA version. That makes the project nicer for people who use IDEs because the header files are linked in the set of target sources.

PR #84, opened 2024-03-08
Added cmath to `half` test and fixed arg mismatch in `cuda_eval`

DoeringChristian:
The function `std::isnan` is provided by the `cmath` library which was not included in the `half` test.
There was a format mismatch in `cuda_eval` causing a runtime exception on the gather test.


wjakob:
Thanks!

PR #86, opened 2024-03-25
Add `jit_registry_peek`

njroussel:
This PR adds an Interface to get an arbitrary pointer to an instance in a given domain: `jit_registry_peek`. This is useful for vectorized dispatches where it is sometimes necessary to trace some function for an arbitrary instance.

The proposed implementation is effectively a compostion of `jit_registry_ptr` and `jit_registry_id_bound`. Returning the highest ID in a domain isn't a particularly desirable property for this function, but it is the only way (I believe) to return a valid pointer in O(1) without making any additional changes to the registry data structures.


wjakob:
Looks great!

PR #88, opened 2024-04-18
Fixes to `jitc_var_reduce_dot` and `jitc_var_call_analyze`

njroussel:
This PR introduces two fixes:

* 0419185c67427dcbc2dc4232ec8e8c6e61eacd8f is a fix for yet another case where we might have used an invalidated `Variable*`
* 4623c62d4e2e7b75d7aecaaa016d7211ec7e92be is a fix for vectorized calls, where we previously might not have completely filled their data map if a callable contained a loop.


wjakob:
LGTM, good catches!

PR #92, opened 2024-07-04
jit_var_migrate: don't copy dependencies of source variable

merlinND:
When the result of `jit_var_migrate` is a new (evaluated) variable, we don't want to copy over the dependencies of the source variable.
It looks to me like they were copied accidentally, since the ref count of the dependencies was not increased.

This issue came up when doing:

```py
a = mi.Float([1, 2, 3])
b = dr.reshape(mi.Float, a, 2, shrink=True)
c = b.numpy()
```

When `c` gets de-allocated, it tried to decrease the dependencies of `b`, which have already been freed.


wjakob:
Thank you Merlin!

PR #93, opened 2024-07-09
Fix typo in fatal error message

Speierers:
This patch fixes a typo in the fatal error message of Dr.Jit core.


njroussel (approved):
Thanks!

PR #99, opened 2024-09-17
Added jit_var_loop_update_inner_in to fix aliasing case

DoeringChristian:
This is the required function for [282](https://github.com/mitsuba-renderer/drjit/pull/282/)


wjakob:
FYI, this commit introduced a reference leak fixed in commit 46bf1f05bfc7. Is it possible that you have reference warnings turned off on your end?

PR #101, opened 2024-09-25
Frozen ThreadState prep

DoeringChristian:
This PR makes the following changes in order to prepare for frozen functions.
- A `KernelFreezing` and `FreezingScope` flags, and associated tests
- A `jit_var_is_unaligned` function, querying if a variable is unaligned
- Adds kernel_param_ids, kernel hash and kernel key to the `ThreadState::launch` funciton interface
- Moves the code, passing kernel params through global memory on CUDA to the `CudaThreadState`
- Adds a `barrier` function and implements it for `LLVM` that simplifies managing concurrency between `LLVM` kernels
- Adds a `notify_free` function to the ThreadState
- Removes the use of `cuMemcpyAsync` in `jitc_var_mem_copy`

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1210,6 +1210,9 @@ extern JIT_EXPORT int jit_var_is_zero_literal(uint32_t index);
 /// Check if a variable represents a normal (not NaN/infinity) literal
 extern JIT_EXPORT int jit_var_is_finite_literal(uint32_t index);
 
+/// Check if the data field of a variable is unaligned
+extern JIT_EXPORT int jit_var_is_unaligned(uint32_t index);
+
 /**
  * \brief Resize a scalar variable to a new size
  *
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1533,13 +1533,20 @@ enum class JitFlag : uint32_t {
     /// is managed automatically and should not be set by application code.
     SymbolicScope = 1 << 18,
 
+    /// Freeze functions annotated with dr.freeze
+    KernelFreezing = 1 << 19,
+
+    /// Set to \c true when Dr.Jit is recording a frozen function
+    FreezingScope = 1 << 20,
+
     /// Default flags
     Default = (uint32_t) ConstantPropagation | (uint32_t) ValueNumbering |
               (uint32_t) FastMath | (uint32_t) SymbolicLoops |
               (uint32_t) OptimizeLoops | (uint32_t) SymbolicCalls |
               (uint32_t) MergeFunctions | (uint32_t) OptimizeCalls |
               (uint32_t) SymbolicConditionals | (uint32_t) ReuseIndices |
-              (uint32_t) ScatterReduceLocal | (uint32_t) PacketOps,
+              (uint32_t) ScatterReduceLocal | (uint32_t) PacketOps |
+              (uint32_t) KernelFreezing,
 
     // Deprecated aliases, will be removed in a future version of Dr.Jit
     LoopRecord = SymbolicLoops,
@@ -1570,6 +1577,8 @@ enum JitFlag {
     JitFlagLaunchBlocking = 1 << 16,
     JitFlagScatterReduceLocal = 1 << 17,
     JitFlagSymbolic = 1 << 18
+    KernelFreezing = 1 << 19,
+    FreezingScope = 1 << 20,
 };
 #endif
 
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1210,7 +1210,12 @@ extern JIT_EXPORT int jit_var_is_zero_literal(uint32_t index);
 /// Check if a variable represents a normal (not NaN/infinity) literal
 extern JIT_EXPORT int jit_var_is_finite_literal(uint32_t index);
 
-/// Check if the data field of a variable is unaligned
+/**
+ * \breif Check if the data field of a variable is unaligned
+ *
+ * This function returns true if the data pointer of the variable is not
+ * properly aligned in memory. Otherwise it returns false.
+ */
 extern JIT_EXPORT int jit_var_is_unaligned(uint32_t index);
 
 /**
--- a/src/call.h
+++ b/src/call.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include "internal.h"
 
 /// Encodes information about a virtual function call
@@ -75,6 +77,8 @@ struct CallData {
         checkpoints.clear();
     }
 };
+
+extern std::vector<CallData *> calls_assembled;
 
 extern uint32_t jitc_var_loop_init(uint32_t *indices, uint32_t n_indices);
 
--- a/src/cuda_ts.h
+++ b/src/cuda_ts.h
@@ -64,5 +64,7 @@ struct CUDAThreadState: ThreadState{
       jitc_raise("jitc_reduce_expanded(): unsupported by CUDAThreadState!");
     }
 
+    void notify_free(const void *) override {};
+
     ~CUDAThreadState(){}
 };
--- a/src/cuda_ts.h
+++ b/src/cuda_ts.h
@@ -3,10 +3,9 @@
 
 struct CUDAThreadState: ThreadState{
 
-    Task *launch(Kernel kernel, uint32_t size,
-                 std::vector<void *> *kernel_params,
-                 uint32_t kernel_param_count,
-                 const uint8_t *kernel_params_global) override;
+    Task *launch(Kernel kernel, KernelKey *key, XXH128_hash_t hash,
+                 uint32_t size, std::vector<void *> *kernel_params,
+                 const std::vector<uint32_t> *) override;
 
     /// Fill a device memory region with constants of a given type
     void memset_async(void *ptr, uint32_t size, uint32_t isize,
--- a/src/cuda_ts.h
+++ b/src/cuda_ts.h
@@ -3,6 +3,8 @@
 
 struct CUDAThreadState: ThreadState{
 
+    void barrier() override{}
+
     Task *launch(Kernel kernel, KernelKey *key, XXH128_hash_t hash,
                  uint32_t size, std::vector<void *> *kernel_params,
                  const std::vector<uint32_t> *) override;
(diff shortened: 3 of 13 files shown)


DoeringChristian:
@wjakob, @merlinND this PR should be ready for review.

src/cuda_ts.h
```diff
@@ -3,10 +3,11 @@
...
 
-    Task *launch(Kernel kernel, uint32_t size,
-                 std::vector<void *> *kernel_params,
-                 uint32_t kernel_param_count,
-                 const uint8_t *kernel_params_global) override;
+    void barrier() override{}
+
+    Task *launch(Kernel kernel, KernelKey *key, XXH128_hash_t hash,
+                 uint32_t size, std::vector<void *> *kernel_params,
+                 const std::vector<uint32_t> *) override;
```
merlinND, line 10: Name of the last argument is missing.
You can comment it `/* kernel_params_global */` if you want to indicate it's unused, but typically I haven't seen that done in method declarations.

src/cuda_ts.h
```diff
@@ -3,10 +3,11 @@
 
 struct CUDAThreadState: ThreadState{
 
-    Task *launch(Kernel kernel, uint32_t size,
-                 std::vector<void *> *kernel_params,
-                 uint32_t kernel_param_count,
-                 const uint8_t *kernel_params_global) override;
+    void barrier() override{}
```
merlinND, line 6: Formatting

src/internal.h
```diff
@@ -652,10 +653,12 @@ struct ThreadState {
     ThreadState() = default;
     ThreadState(const ThreadState &other) = default;
 
-    virtual Task *launch(Kernel kernel, uint32_t size,
-                         std::vector<void *> *kernel_params,
-                         uint32_t kernel_param_count,
-                         const uint8_t *kernel_params_global) = 0;
+    /// Inserts a barrier task
```
merlinND, line 656: Expand a little bit on what this means, whether all backends are supported or not, etc.

src/call.h
```diff
@@ -76,6 +78,8 @@ struct CallData {
     }
 };
 
+extern std::vector<CallData *> calls_assembled;
```
merlinND, line 81: Does `calls_assembled` need to be exposed to all people who include `call.h`? (Fine if it does, just checking)

src/internal.h
```diff
@@ -711,6 +714,8 @@ struct ThreadState {
     /// dr.ReduceOp.Expand
     virtual void reduce_expanded(VarType vt, ReduceOp op, void *data,
                                  uint32_t exp, uint32_t size) = 0;
+
+    virtual void notify_free(const void *ptr) = 0;
```
merlinND, line 718: Docstring, explain what it does and that it's needed for kernel freezing.

src/op.cpp
```diff
@@ -1789,7 +1789,9 @@ uint32_t jitc_var_gather(uint32_t src_, uint32_t index, uint32_t mask) {
         jitc_var_eval(src);
 
     /// Perform a memcpy when this is a size-1 literal load
-    if (!result && var_info.size == 1 && var_info.literal) {
+    /// (if not freezing a function)
```
merlinND, line 1792: If possible, explain in 1-2 sentences why this optimization is skipped during kernel recording.

src/var.cpp
```diff
@@ -1436,6 +1436,10 @@ void jitc_var_read(uint32_t index, size_t offset, void *dst) {
     if (v->is_literal() || v->is_undefined()) {
         memcpy(dst, &v->literal, isize);
     } else if (v->is_evaluated()) {
+        if (jitc_flags() & (uint32_t) JitFlag::FreezingScope)
+            jitc_raise("jit_var_read(): reading from evaluated variables while "
```
merlinND, line 1440: @wjakob Under the new policy of never crashing the kernel and including more troubleshooting info, how could we make it easy for the user to find the source of this illegal `jit_var_read()`?
Would something like "enable the Debug flag and we will show the Python line that triggers this" be possible?

src/llvm_ts.cpp
```diff
@@ -287,10 +286,31 @@ static void submit_cpu(KernelType type, Func &&func, uint32_t width,
...
+                                         (uint32_t) scheduled_tasks.size());
+        task_release(jitc_task);
+        for (Task *t : scheduled_tasks)
+            task_release(t);
+        jitc_task = new_task;
+    }
+    scheduled_tasks.clear();
+}
+
+Task *LLVMThreadState::launch(Kernel kernel, KernelKey *, XXH128_hash_t,
```
merlinND, line 311: I like the convention of putting the param name in an inline comment if it's unused, though I don't know if that's used consistently in this codebase:
```cpp
Kernel kernel, KernelKey * /*key*/, ...
```

src/llvm_ts.cpp
```diff
@@ -287,10 +286,31 @@ static void submit_cpu(KernelType type, Func &&func, uint32_t width,
...
-Task *LLVMThreadState::launch(Kernel kernel, uint32_t size,
-                              std::vector<void *> *kernel_params,
-                              uint32_t /* kernel_param_count */,
-                              const uint8_t* /* kernel_params_global */) {
+/// Temporary scratch space for scheduled tasks
+static std::vector<Task *> scheduled_tasks;
+
+void LLVMThreadState::barrier() {
+    if (scheduled_tasks.size() == 1) {
+        task_release(jitc_task);
```
merlinND, line 294: I don't know about the LLVM backends and the task system, but is this safe to do at any point?
If not, the docstring of `ThreadState::barrier()` should specify exactly when it's safe & intended to call this function.

src/llvm_ts.cpp
```diff
@@ -287,10 +286,31 @@ static void submit_cpu(KernelType type, Func &&func, uint32_t width,
...
+        jitc_task = scheduled_tasks[0];
+    } else {
+        jitc_assert(!scheduled_tasks.empty(),
+                    "jit_eval(): no tasks generated!");
+
+        // Insert a barrier task
+        Task *new_task = task_submit_dep(nullptr, scheduled_tasks.data(),
+                                         (uint32_t) scheduled_tasks.size());
+        task_release(jitc_task);
```
merlinND, line 303: Is it always the case that `jitc_task == scheduled_task.front()` (or maybe `.back()`)?
In this function, I feel like either `jitc_task` is being freed twice, or it's being dropped (= not included in the dependencies of `new_task`).
But I'm not familiar with the task system.

src/eval.cpp
```diff
@@ -733,36 +718,14 @@ void jitc_eval_impl(ThreadState *ts) {
...
-            // Insert a barrier task
-            Task *new_task = task_submit_dep(nullptr, scheduled_tasks.data(),
-                                             (uint32_t) scheduled_tasks.size());
-            task_release(jitc_task);
-            for (Task *t : scheduled_tasks)
-                task_release(t);
-            jitc_task = new_task;
-        }
-    }
+    ts->barrier();
```
merlinND, line 728: Explain in a few words what this barrier achieves / why it's needed. E.g. ~"Ensure all kernels launched here finish before subsequent ones can start" (if I understood correctly).

merlinND (commented):
Very nice, thanks for making this easy to review!

src/call.h
DoeringChristian: Yes, I need it to capture offset buffers. Should I document that?

include/drjit-core/jit.h
```diff
@@ -1210,6 +1210,9 @@ extern JIT_EXPORT int jit_var_is_zero_literal(uint32_t index);
 /// Check if a variable represents a normal (not NaN/infinity) literal
 extern JIT_EXPORT int jit_var_is_finite_literal(uint32_t index);
 
+/// Check if the data field of a variable is unaligned
```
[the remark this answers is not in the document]
DoeringChristian: Right, I'll remove the LLVM requirement (my mistake) and add documentation.

src/llvm_ts.cpp
DoeringChristian: Good point.
To me it seems fine to call it at any point, but I'm also not 100% sure.

src/call.h
merlinND: No I think that's alright, I was just checking :+1:

src/var.cpp
wjakob: The exception will bubble up through the C++ code and then be re-raised from Python. As far as I can see, this is already good enough.

src/llvm_ts.cpp
wjakob: I guess it just makes sense to call it after a set of kernel launches.  All the other operations have implicit barriers between them.

src/eval.cpp
wjakob: Suggest rephrasing: "The barrier ensures that subsequently launched kernel can't start to run until all kernels in the current launch have finished."

src/internal.h
```diff
@@ -652,10 +653,16 @@ struct ThreadState {
     ThreadState() = default;
     ThreadState(const ThreadState &other) = default;
 
-    virtual Task *launch(Kernel kernel, uint32_t size,
-                         std::vector<void *> *kernel_params,
-                         uint32_t kernel_param_count,
-                         const uint8_t *kernel_params_global) = 0;
+    /// Schedules a barrier taks.
+    /// Ensures that all kernel kernel launches are executed after all tasks
```
wjakob, line 657: ditto

src/llvm_ts.cpp
wjakob: How about: "All tasks in 'scheduled_tasks' have 'jitc_task' (if present) as parent. To create a barrier, we release 'jitc_task' and initialize it with a new dummy task that has all members of 'scheduled_tasks' as parents. This allows groups of kernel launches to run in parallel, while serializing subsequent groups."

njroussel (approved):
This looks great! I believe all open discussions have been addressed, I don't have anything to add.

@DoeringChristian Could you please squash the "review" commits? We can merge this PR once that's done :rocket:

DoeringChristian:
> This looks great! I believe all open discussions have been addressed, I don't have anything to add.
> 
> @DoeringChristian Could you please squash the "review" commits? We can merge this PR once that's done 🚀

Should I squash them into a single review commit or squash them into the previous commits?

wjakob:
(I will do the squashing myself in the interest of time, I wanted to rebase another commit on top of these changes.)

PR #105, opened 2024-10-18
Fix support for disabling ray tracing op in OptiX

dvicini:
This partially addresses the issue https://github.com/mitsuba-renderer/mitsuba3/issues/1349 in Mitsuba.

Currently, invalid PTX is generated if a ray tracing op is intended to be disabled. This PR adds the missing label for the generated code to be valid.

Alternatively, we could also just disable this jump if ray tracing is anyway disabled. Not sure what's the cost of the jump vs setting a number of registers to 0.


njroussel (commented):
Hi @dvicini 

Thanks for reporting this, I've fixed it by completely removing the branching in https://github.com/mitsuba-renderer/drjit-core/commit/fc80ebd30aa5b6c894235163f9d0979ec7a3f3d4.

PR #106, opened 2024-10-23
All domain Registry Traversal per Backend

DoeringChristian:
The PR adds the option to traverse all registry entries for a backend.

This is required for function freezing, as it is not easily possible to get the domains of a variable representing a pointer, when traversing a C++ class such as a scene.
The fix is to just traverse the whole registry.

This PR makes the following changes:
- modifies the `jitc_registry_id_bound` function to return the bound over all domains in the backend if the `domain` argument is a `nullptr`
- Adds a `jitc_registry_fill_ptrs` function, taking a pointer to a memory region to fill with the registered pointers.
- Adds the backend as a field to the `Domain` struct

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -494,13 +494,14 @@ extern JIT_EXPORT void jit_registry_remove(const void *ptr);
 extern JIT_EXPORT uint32_t jit_registry_id(const void *ptr);
 
 /// Return the largest instance ID for the given domain
-/// If the \c domain is a nullptr, it returns the largest instance ID for the given backend
+/// If the \c domain is a nullptr, it returns the number of active entries in
+/// all domains for the given backend
 extern JIT_EXPORT uint32_t jit_registry_id_bound(JitBackend backend,
                                                  const char *domain);
 
 /// Fills the \c dest pointer array with all pointers registered in the registry
 /// \c dest has to point to an array with \c jit_registry_id_bound(backend, nullptr) entries
-extern JIT_EXPORT void jit_registry_fill_ptrs(JitBackend backend, void **dest);
+extern JIT_EXPORT void jit_registry_get_pointers(JitBackend backend, void **dest);
 
 /// Return the pointer value associated with a given instance ID
 extern JIT_EXPORT void *jit_registry_ptr(JitBackend backend,
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -494,8 +494,13 @@ extern JIT_EXPORT void jit_registry_remove(const void *ptr);
 extern JIT_EXPORT uint32_t jit_registry_id(const void *ptr);
 
 /// Return the largest instance ID for the given domain
+/// If the \c domain is a nullptr, it returns the largest instance ID for the given backend
 extern JIT_EXPORT uint32_t jit_registry_id_bound(JitBackend backend,
                                                  const char *domain);
+
+/// Fills the \c dest pointer array with all pointers registered in the registry
+/// \c dest has to point to an array with \c jit_registry_id_bound(backend, nullptr) entries
+extern JIT_EXPORT void jit_registry_fill_ptrs(JitBackend backend, void **dest);
 
 /// Return the pointer value associated with a given instance ID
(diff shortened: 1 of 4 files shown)


DoeringChristian:
@wjakob, @merlinND This PR should be ready for review.

include/drjit-core/jit.h
```diff
@@ -494,9 +494,14 @@ extern JIT_EXPORT void jit_registry_remove(const void *ptr);
 extern JIT_EXPORT uint32_t jit_registry_id(const void *ptr);
 
 /// Return the largest instance ID for the given domain
+/// If the \c domain is a nullptr, it returns the largest instance ID for the given backend
 extern JIT_EXPORT uint32_t jit_registry_id_bound(JitBackend backend,
                                                  const char *domain);
 
+/// Fills the \c dest pointer array with all pointers registered in the registry
+/// \c dest has to point to an array with \c jit_registry_id_bound(backend, nullptr) entries
+extern JIT_EXPORT void jit_registry_fill_ptrs(JitBackend backend, void **dest);
```
merlinND, line 503: "fill_ptrs" could sound like you're filling the registry. Maybe `jit_registry_all_pointers()` or `jit_registry_get_pointers()`?

include/drjit-core/jit.h
```diff
@@ -494,9 +494,14 @@ extern JIT_EXPORT void jit_registry_remove(const void *ptr);
 extern JIT_EXPORT uint32_t jit_registry_id(const void *ptr);
 
 /// Return the largest instance ID for the given domain
+/// If the \c domain is a nullptr, it returns the largest instance ID for the given backend
```
merlinND, line 497: Actually it looks like it just counts the number of *active* entries in a domain, not the largest instance ID?

src/registry.cpp
```diff
@@ -149,13 +151,38 @@ uint32_t jitc_registry_id(const void *ptr) {
 
 uint32_t jitc_registry_id_bound(JitBackend backend, const char *domain) {
     Registry &r = registry;
+    if (!domain) {
+        uint32_t i = 0;
+        for (Domain &domain : r.domains) {
+            if (domain.backend == backend)
```
merlinND, line 157: I guess you could create a `DomainKey{ backend, domain_name }` to look up the domain ID into `r.domain_ids`, but I don't think it would be faster than your current implementation.

src/registry.cpp
```diff
@@ -149,13 +151,38 @@ uint32_t jitc_registry_id(const void *ptr) {
 
 uint32_t jitc_registry_id_bound(JitBackend backend, const char *domain) {
     Registry &r = registry;
+    if (!domain) {
+        uint32_t i = 0;
```
merlinND, line 155:
```suggestion
        uint32_t n = 0;
```
since it's a count and not an index.

src/registry.cpp
```diff
@@ -149,13 +151,38 @@ uint32_t jitc_registry_id(const void *ptr) {
...
+    }
     auto it = r.domain_ids.find(DomainKey{ backend, domain });
     if (it == r.domain_ids.end())
         return 0;
     else
         return r.domains[it->second].id_bound;
 }
 
+void jitc_registry_fill_ptrs(JitBackend backend, void **dest) {
+    Registry &r = registry;
```
merlinND, line 172:
```suggestion
    const Registry &r = registry;
```

src/registry.h
```diff
@@ -22,8 +22,13 @@ extern void jitc_registry_remove(const void *ptr);
 extern uint32_t jitc_registry_id(const void *ptr);
 
 /// Return the largest instance ID for the given domain
+/// If the \c domain is a nullptr, it returns the largest instance ID for the given backend
```
merlinND, line 25: Same comment as above

merlinND (commented):
Thanks for making this one easy to review as well!

src/registry.cpp
DoeringChristian: Agreed. I guess it depends on how many domains there are with a different backend.

wjakob (approved):

wjakob:
Looks good!

src/registry.cpp
DoeringChristian: I'll leave it like this for now.

PR #107, opened 2024-10-23
Frozen ThreadState

DoeringChristian:
This PR implements the Freezing feature in `drjit-core` (depends on  #106).

It makes the following changes:
- Adds the freezing api to `jit.h`, including the following functions: `jit_freeze_start`, `jit_freeze_stop`, `jit_freeze_replay`, `jit_freeze_dry_run`, `jit_freeze_pause`, `jit_freeze_resume`, `jit_freeze_abort` and `jit_freeze_destroy` as well as the `Recording` struct.
- Adds the implementation of the `RecordThreadState`, a wrapper around either the CUDA or LLVM `ThreadState` for freezing them.

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -2483,6 +2483,154 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
 extern JIT_EXPORT uint32_t jit_array_read(uint32_t source, uint32_t offset,
                                           uint32_t mask);
 
+// ====================================================================
+//     Core functionality to enable kernel freezing via @dr.freeze
+// ====================================================================
+
+/// Opaque data structure, storing a sequence of Dr.Jit operations recorded
+/// using the API below.
+struct Recording;
+
+/**
+ * \brief Start a recording session. This causes Dr.Jit to track all backend
+ * operations such as memory copies and kernel launches, storing them into a
+ * representation that can be replayed without needing to re-trace code.
+ *
+ * \param backend
+ *     The backend for which recording should be started.
+ *
+ * \param inputs
+ *     An array of input variable indices, which have to be specified
+ *     when starting the recording. Different indices, representing other
+ *     variables, might be provided when replaying. This function borrows the
+ *     indices and does not increment their refcount.
+ *
+ * \param n_inputs
+ *      The number of input variables for the recording
+ */
+extern JIT_EXPORT void
+jit_freeze_start(JitBackend backend, const uint32_t *inputs, uint32_t n_inputs);
+
+/**
+ * \brief Stop recording operations and return a struct containing the
+ * recording.
+ *
+ * The recording is returned as an opaque pointer and has to be destroyed
+ * afterwards by calling the \c jitc_freeze_destroy function.
+ *
+ * \param backend
+ *     The backend on which recording should be stopped.
+ *
+ * \param outputs
+ *     An array of output variable indices. When replaying the recording, these
+ *     variables are returned from the replay function. This function borrows
+ *     the indices and does not modify their refcount.
+ *
+ * \param n_outputs
+ *     The number of output variables of the recording.
+ */
+extern JIT_EXPORT Recording *jit_freeze_stop(JitBackend backend,
+                                             const uint32_t *outputs,
+                                             uint32_t n_outputs);
+
+/**
+ * \brief Replay a recording with different inputs.
+ *
+ * Replaying a recording with different inputs results in different output
+ * variables. Their variable indices will be written into the outputs array.
+ *
+ * \param recording
+ *     The recording to replay given different inputs.
+ *
+ * \param inputs
+ *     An array of input variable indices for replaying the recording. The
+ *     number of inputs taken from the array is equal to the number of inputs
+ *     supplied to the jit_start_record function. The variable indices are only
+ *     referenced and their refcount is not changed by this function.
+ *
+ * \param outputs
+ *     This array is filled with the output variable indices, created when
+ *     replaying the recording. The size of the array has to be equal to the
+ *     number of output variables supplied to the jit_record_stop function.
+ *     The variables in the output are borrowing references and have to be
+ *     released.
+ */
+extern JIT_EXPORT void jit_freeze_replay(Recording *recording,
+                                         const uint32_t *inputs,
+                                         uint32_t *outputs);
+
+/**
+ * \brief Perform a dry run replay of a recording (if required), which does not
+ * perform any actual work.
+ *
+ * A dry run of the recording calculates the widths of all kernels and returns
+ * ``0`` if the function has to be re-recorded. It returns ``1`` if the function
+ * can be replayed. This is required to catch cases where the size of an input
+ * variable changes the compiled kernels, for example when performing scatter
+ * reductions in LLVM mode.
+ *
+ * \param recording
+ *     The recording to replay given different inputs.
+ *
+ * \param inputs
+ *     An array of input variable indices for replaying the recording. The
+ *     number of inputs taken from the array is equal to the number of inputs
+ *     supplied to the jit_start_record function. No actual changes are
+ *     performed on the variables in dry-run mode.
+ *
+ * \return 0 if retracing the function is required, 1 otherwise
+ */
+extern JIT_EXPORT int jit_freeze_dry_run(Recording *recording,
+                                         const uint32_t *inputs);
+
+/**
+ * \brief Pause recording the ThreadState for this backend.
+ *
+ * Returns an integer indicating the pause state before calling this function,
+ * i.e. returns ``1`` if recording was already paused, and ``0`` otherwise. If
+ * no recording is in progress, this function fails. This is useful to prevent
+ * accidentally recording operations when traversing the output of a frozen
+ * function.
+ *
+ * \param backend
+ *     The backend for which to pause recording the thread state.
+ */
+extern JIT_EXPORT int jit_freeze_pause(JitBackend backend);
+
+/**
+ * \brief Resume recording the ThreadState for this backend.
+ *
+ * Returns an integer indicating the pause state before calling this function,
+ * i.e. returns ``1`` if recording was paused, and ``0`` otherwise. If no
+ * recording is in progress, this function fails.
+ *
+ * \param backend
+ *     The backend for which to pause recording the thread state.
+ */
+extern JIT_EXPORT int jit_freeze_resume(JitBackend backend);
+
+/**
+ * \brief Abort recording the ThreadState for this backend.
+ *
+ * This will abort the recording process and restore the state to the state it
+ * was in before starting the recording. Aborting a recording has the same
+ * effect as never starting the recording. All operations on variables are still
+ * performed and their state might have changed. If no recording is in progress,
+ * this function will run without effect.
+ *
+ * \param backend
+ *     The backend for which to abort recording the thread state.
+ */
+extern JIT_EXPORT void jit_freeze_abort(JitBackend backend);
+
+/**
+ * \brief Destroys a recording and frees the associated memory.
+ *
+ * \param recording
+ *     The recording to destroy.
+ */
+extern JIT_EXPORT void jit_freeze_destroy(Recording *recording);
+
 #if defined(__cplusplus)
 }
 #endif
--- a/src/record_ts.cpp
+++ b/src/record_ts.cpp
@@ -1,0 +1,2204 @@
+/** Frozen function backend
+ *
+ * This file implements the backend for the `FrozenFunction` feature.
+ * Calling ``jitc_freeze_start`` allows recording of ``ThreadState`` operations
+ * by swapping out the current thread state with a ``RecordThreadState``, which
+ * records all operations performed on itself. Ending the recording process by
+ * calling `jit_freeze_stop`, produces a ``Recording``, which can be replayed
+ * later.
+ *
+ * The ``Recording`` relies on three main data structures:
+ *
+ * - ``recording.recorded_variables``: A vector of variables that have been
+ *   encountered when recording. We track variables by the memory regions they
+ *   refer to, therefore only evaluated variables can be tracked. During
+ *   recording, we map the pointers to their ``RecordedVariables`` using the
+ *   ``ptr_to_slot`` hash map. For that reason pointers have to refer to the
+ *   start of a memory region, allocated with ``jic_malloc``.
+ *
+ * - ``recording.operations``: A vector of operations, that have been recorded.
+ *   Each operation stores its type, some additional information, such as the
+ *   executed kernel, and a tuple of indices into the ``recording.dependencies``
+ *   vector. This records the accesses the operation has performed on variables.
+ *
+ * - ``recording.dependencies``: A vector of AccessInfo which are used used to
+ *   record how operations access different variables. Every entry records an
+ *   index (``slot``) into the ``recording.recorded_variables`` vector,
+ *   representing the memory region it refers to. Additionally, it stores the
+ *   type of access (Input or Output) and the variable type as which the
+ *   variable was accessed. The latter is required since some operations might
+ *   interpret the same memory with different types. For example, a ``memcpy``
+ *   operation might access the an ``UInt32`` array as an array of ``UInt8``
+ *   bytes.
+ *
+ * To identify memory regions by their pointers and map them to indices in
+ * ``recording.recorded_variables``, we employ the ``ptr_to_slot`` hash map.
+ *
+ * When recording a new operation, it is pushed on the back of the
+ * ``recording.operations`` array. If the operation accesses a variable that
+ * already exists i.e. it is the result of another operation or the input of the
+ * Frozen Function, ``get_variable`` should be used to lookup the index into the
+ * ``recorded_variables``, representing the variable. The function will throw an
+ * exception if the pointer has not yet been encountered.
+ * On the other hand, if the variable is either a result of the operation or it
+ * is accessed by that operation for the first time (see aggregate),
+ * ``add_variable`` should be used. This will either push a new
+ * ``RecordedVariable`` to the ``recorded_variables`` vector and insert an index
+ * to it into ``ptr_to_slot`` or lookup the one already known to
+ * ``ptr_to_slot``. The index returned by either function can be used to record
+ * how the memory region/variable is accessed by the operation. To this end, a
+ * ``AccessInfo`` struct is constructed, with at least its ``slot`` being set to
+ * the recorded variable index and its `type` to either ``ParamType::Input`` or
+ * ``ParamType::Output``. Using ``add_param`` the struct is then pushed onto the
+ * ``dependencies`` vector, while updating the variable type and state of the
+ * referenced ``RecordedVariable``. We provide wrapper functions for recording
+ * both input and output accesses to variables, which take the pointer address
+ * and an optional variable type.
+ *
+ * Example in aggregate operation:
+ * ```
+ * uint32_t slot = add_variable(p.src);
+ *
+ * AccessInfo info;
+ * info.slot           = slot;
+ * info.type           = ParamType::Input;
+ * info.pointer_access = p.size == 8;
+ * info.extra.offset   = p.offset;
+ * info.test_uninit    = false;
+ * add_param(info);
+ * ```
+ *
+ * Example of recording `reduce_dot`:
+ * This code is executed in the `record_reduce_dot` function and shows how to
+ * record an operation. The functions ``add_out_param``, ``add_in_param`` as
+ * well as the fields ``m_recording`` are part of the ``RecordThreadState``.
+ * ```
+ * uint32_t start = m_recording.dependencies.size();
+ * add_out_param(out, type);
+ * add_in_param(ptr_1, type);
+ * add_in_param(ptr_2, type);
+ * uint32_t end = m_recording.dependencies.size();
+ *
+ * Operation op;
+ * op.type             = OpType::ReduceDot;
+ * op.dependency_range = std::pair(start, end);
+ * op.size             = size;
+ * m_recording.operations.push_back(op);
+ * ```
+ * In order to replay the recorded ThreadState, we have to first call the
+ * ``jit_freeze_dry_run`` function. This will determine if the recording has to
+ * be re-traced, by replaying the recording without executing the operations or
+ * allocating memory. If that function returns with ``1``, the
+ * ``jitc_freeze_replay`` function can be called to actually replay the
+ * recording.
+ *
+ *
+ *
+ * The following is a list of complications, that might occur when trying to
+ * record frozen functions:
+ * - Since we allow replaying frozen functions with differently sized input
+ *   variables, changes to the width will not be tracked by the function
+ *   interface key. In order to still allow using the width of one variable to
+ *   define the width of another, for example,
+ *   ``dr.gather(t, x, dr.width(x))``, we implement a heuristic that determines
+ *   the kernel launch size dependent on the sizes of its inputs. The heuristic
+ *   might fall back to the recorded kernel size, which might not be correct.
+ *   Introducing a ``symbolic_width`` function can resolve this.
+ * - Some operations might require dry running the recording before being able
+ *   to replay them. One example, is the ``reduce_expanded`` case, which results
+ *   in different kernels depending on the size of its input size. During dry
+ *   running, the size of variable is inferred in the same way as when
+ *   replaying, without actually executing any operations. In these cases, we
+ *   set the ``requires_dry_run`` flag in the recording. The ``Compress``
+ *   operation however creates an output which size depends on the content of
+ *   the compressed array. We therefore return false when encountering a
+ *   ``Comrpess`` operation in dry run mode, which should result in a
+ *   re-recording of the frozen function.
+ * - Recording the ``poke`` operation is currently unsupported, as it will write
+ *   to some offset pointer in a memory region, which cannot be tracked.
+ * - Exceptions that occur in the ``record_*`` functions are captured instead of
+ *   propagated to the calling function. Letting the exception propagate from a
+ *   function such as \ref launch will leak variables, since they are still held
+ *   by the \ref schedule. Therefore, all other operations might still be
+ *   executed and the content of the variables will have changed.
+ */
+
+#include "record_ts.h"
+#include "call.h"
+#include "common.h"
+#include "drjit-core/jit.h"
+#include "eval.h"
+#include "internal.h"
+#include "llvm.h"
+#include "log.h"
+#include "malloc.h"
+#include "profile.h"
+#include "util.h"
+#include "var.h"
+
+const char *op_type_name[(int) OpType::Count]{
+    "Barrier",        "KernelLaunch",      "MemsetAsync", "Expand",
+    "ReduceExpanded", "Compress",          "MemcpyAsync", "Mkperm",
+    "BlockReduce",    "BlockPrefixReduce", "ReduceDot",   "Aggregate",
+    "Free",
+};
+
+static bool dry_run = false;
+
+/**
+ * Represents a variable during replay.
+ * It is created from the RecordVariable at the top of the replay function.
+ */
+struct ReplayVariable {
+    void *data = nullptr;
+    // Tracks the capacity, this allocation has been allocated for
+    size_t alloc_size = 0;
+    // Tracks the size in bytes, of this allocation
+    size_t data_size = 0;
+    uint32_t index;
+    RecordedVarInit init;
+
+    ReplayVariable(RecordedVariable &rv) : index(rv.index), init(rv.init) {
+        if (init == RecordedVarInit::Captured) {
+            // Copy the variable, so that it isn't changed by this recording.
+            // Captured variables are only used for vcall offset buffers, which
+            // are not supposed to change between replay calls.
+            index = jitc_var_copy(index);
+
+            Variable *v = jitc_var(index);
+            data        = v->data;
+            alloc_size  = v->size * type_size[v->type];
+            data_size   = alloc_size;
+        }
+    }
+
+    /// Initializes the \ref ReplayVariable from a function input.
+    void init_from_input(Variable *input_variable) {
+        data       = input_variable->data;
+        alloc_size = type_size[input_variable->type] * input_variable->size;
+        data_size = alloc_size;
+    }
+
+    /**
+     * Calculate the number of elements from the size in bytes given some
+     * variable type.
+     */
+    uint32_t size(VarType vtype) { return size(type_size[(uint32_t) vtype]); }
+    /**
+     * Calculate the number of elements given some type size.
+     * This depends on how an operation accesses a variable.
+     * For example, a memcpy operation might access a variable as an array of
+     * \ref uint8_t types, whereas a kernel can access the same variable as an
+     * array of ``uint64_t``. This changes the size of the variable when
+     * inferring the size of the kernel launch.
+     */
+    uint32_t size(uint32_t tsize) {
+        if (tsize == 0)
+            jitc_raise("replay(): Invalid variable type!");
+        size_t size = (data_size / (size_t) tsize);
+
+        if (size == 0)
+            jitc_raise("replay(): Error, determining size of variable! init "
+                       "%u, dsize=%zu",
+                       (uint32_t) init, data_size);
+
+        if (size * (size_t) tsize != data_size)
+            jitc_raise("replay(): Error, determining size of variable!");
+
+        return (uint32_t) size;
+    }
+
+    void alloc(JitBackend backend, uint32_t size, VarType type) {
+        alloc(backend, size, type_size[(uint32_t) type]);
+    }
+    void alloc(JitBackend backend, uint32_t size, uint32_t isize) {
+        size_t dsize = ((size_t) size) * ((size_t) isize);
+        return alloc(backend, dsize);
+    }
+    /**
+     * Allocates the data for this replay variable. If this is attempted twice,
+     * we test Whether the allocated size is sufficient and re-allocate the
+     * memory if necessary.
+     *
+     * In dry run mode, this does not perform allocations, but just changes the
+     * \c data_size of the \c ReplayVariable, so that subsequent operations can
+     * infer the size of this variable.
+     */
+    void alloc(JitBackend backend, size_t dsize) {
+        AllocType alloc_type =
+            backend == JitBackend::CUDA ? AllocType::Device : AllocType::Host;
+
+        if (!data) {
+            alloc_size = dsize;
+            jitc_log(LogLevel::Debug, "    allocating output of size %zu.",
+                     dsize);
+            if (!dry_run)
+                data = jitc_malloc(alloc_type, dsize);
+        } else if (alloc_size < dsize) {
+            alloc_size = dsize;
+            jitc_log(LogLevel::Debug, "    re-allocating output of size %zu.",
+                     dsize);
+            if (!dry_run) {
+                jitc_free(data);
+                data = jitc_malloc(alloc_type, dsize);
+            }
+        } else {
+            // Do not reallocate if the size is enough
+        }
+
+        data_size = dsize;
+    }
+
+    void free() {
+        jitc_free(data);
+        data       = nullptr;
+        data_size  = 0;
+        alloc_size = 0;
+    }
+};
+
+/// Kernel parameter buffer
+static std::vector<void *> kernel_params;
+
+/// ThreadState used during replay
+static ThreadState *ts = nullptr;
+
+/// Temporary variables used for replaying a recording.
+static std::vector<ReplayVariable> replay_variables;
+
+static ProfilerRegion pr_operation("Replay Operation");
+static ProfilerRegion pr_kernel_launch("KernelLaunch");
+static ProfilerRegion pr_barrier("Barrier");
+static ProfilerRegion pr_memset_async("MemsetAsync");
+static ProfilerRegion pr_reduce_expanded("ReduceExpanded");
+static ProfilerRegion pr_expand("Expand");
+static ProfilerRegion pr_compress("Compress");
+static ProfilerRegion pr_memcpy_async("MemcpyAsync");
+static ProfilerRegion pr_mkperm("Mkperm");
+static ProfilerRegion pr_block_reduce("BlockReduce");
+static ProfilerRegion pr_block_prefix_reduce("BlockPrefixReduce");
+static ProfilerRegion pr_reduce_dot("ReduceDot");
+static ProfilerRegion pr_aggregate("Aggregate");
+static ProfilerRegion pr_free("Free");
+static ProfilerRegion pr_output("Output");
+
+int Recording::replay(const uint32_t *replay_inputs, uint32_t *replay_outputs) {
+
+    ts = thread_state(backend);
+
+#ifndef NDEBUG
+    n_kernels = 0;
+#endif
+
+    if (dynamic_cast<RecordThreadState *>(ts) != nullptr)
+        jitc_raise("replay(): Tried to replay while recording!");
+
+    OptixShaderBindingTable *tmp_sbt = ts->optix_sbt;
+    scoped_set_context_maybe guard2(ts->context);
+
+    // Initialize replay_variables
+    replay_variables.clear();
+    replay_variables.reserve(recorded_variables.size());
+    for (RecordedVariable &rv : recorded_variables) {
+        replay_variables.push_back(ReplayVariable(rv));
+    }
+
+    jitc_log(LogLevel::Debug, "replay(): inputs");
+    // Populate with input variables
+    for (uint32_t i = 0; i < inputs.size(); ++i) {
+        Variable *input_variable = jitc_var(replay_inputs[i]);
+        replay_variables[inputs[i]].init_from_input(input_variable);
+        jitc_log(LogLevel::Debug, "    input %u: r%u maps to slot s%u", i,
+                 replay_inputs[i], inputs[i]);
+    }
+
+    // The main loop that executes each operation
+    for (uint32_t i = 0; i < operations.size(); ++i) {
+        Operation &op = operations[i];
+        ProfilerPhase profiler(pr_operation);
+        if (!op.enabled)
+            continue;
+
+        switch (op.type) {
+            case OpType::KernelLaunch:
+                if (!replay_launch(op))
+                    return false;
+                break;
+            case OpType::Barrier: {
+                ProfilerPhase profiler(pr_barrier);
+                if (!dry_run)
+                    ts->barrier();
+            }; break;
+            case OpType::MemsetAsync:
+                if (!replay_memset_async(op))
+                    return false;
+                break;
+            case OpType::ReduceExpanded:
+                if (!replay_reduce_expanded(op))
+                    return false;
+                break;
+            case OpType::Expand:
+                if (!replay_expand(op))
+                    return false;
+                break;
+            case OpType::Compress:
+                if (!replay_compress(op))
+                    return false;
+                break;
+            case OpType::MemcpyAsync:
+                if (!replay_memcpy_async(op))
+                    return false;
+                break;
+            case OpType::Mkperm:
+                if (!replay_mkperm(op))
+                    return false;
+                break;
+            case OpType::BlockReduce:
+                if (!replay_block_reduce(op))
+                    return false;
+                break;
+            case OpType::BlockPrefixReduce:
+                if (!replay_block_prefix_reduce(op))
+                    return false;
+                break;
+            case OpType::ReduceDot:
+                if (!replay_reduce_dot(op))
+                    return false;
+                break;
+            case OpType::Aggregate:
+                if (!replay_aggregate(op))
+                    return false;
+                break;
+            case OpType::Free: {
+                ProfilerPhase profiler(pr_free);
+
+                uint32_t i         = op.dependency_range.first;
+                AccessInfo info     = dependencies[i];
+                ReplayVariable &rv = replay_variables[info.slot];
+
+                rv.free();
+
+            }; break;
+            default:
+                jitc_fail(
+                    "An unknown operation with type %u has been recorded!",
+                    (uint32_t) op.type);
+                break;
+        }
+    }
+
+    ts->optix_sbt = tmp_sbt;
+
+    if (dry_run)
+        return true;
+
+    ProfilerPhase profiler(pr_output);
+    // Create output variables
+    jitc_log(LogLevel::Debug, "replay(): creating outputs");
+    for (uint32_t i = 0; i < outputs.size(); ++i) {
+        AccessInfo info     = outputs[i];
+        uint32_t slot      = info.slot;
+        ReplayVariable &rv = replay_variables[slot];
+
+        if (rv.init == RecordedVarInit::Input) {
+            // Use input variable
+            jitc_log(LogLevel::Debug,
+                     "    output %u: from slot s%u = input[%u]", i, slot,
+                     rv.index);
+            jitc_assert(rv.data, "replay(): freed an input variable "
+                                 "that is passed through!");
+            uint32_t var_index = replay_inputs[rv.index];
+            jitc_var_inc_ref(var_index);
+            replay_outputs[i] = var_index;
+        } else if (rv.init == RecordedVarInit::Captured) {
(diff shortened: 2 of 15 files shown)


DoeringChristian:
@merlinND, could you please do a preliminary pass over this PR, to catch some obvious mistakes of mine. Note, this also includes the commit from #106, which I'll remove when that gets merged.

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+        UInt32 i0(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+        UInt32 r0 = opaque<UInt32>(45, 1);
+
+        uint32_t inputs[] = {
+            i0.index(),
+        };
+
+        jit_freeze_start(Backend, inputs, 1);
+
+        UInt32 o0 = hsum(i0);
```
merlinND, line 381: It would be more interesting if there was one operation creating a first kernel, and then `hsum()` is applied on the output of the first kernel.

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+    jit_freeze_destroy(recording);
+}
+
+/**
+ * Basic addition test.
+ * Supplying a different input should replay the operation, with this input.
+ * In this case, the input at replay is incremented and should result in an
+ * incremented output.
+ */
+TEST_BOTH(9_resized_input) {
```
merlinND, line 471: Wouldn't be needed anymore if you change the width during replay for all of the previous tests.

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+        jit_freeze_replay(recording, inputs, outputs);
+
+        jit_log(LogLevel::Info, "o0: %s", jit_var_str(outputs[0]));
+        jit_assert(jit_var_all(jit_var_eq(r0.index(), outputs[0])));
+    }
+
+    jit_freeze_destroy(recording);
+}
+
+TEST_BOTH(10_input_passthrough) {
```
merlinND, line 511: Comment describing the test case, like for the previous tests.

merlinND (commented):
Review part 1. It's mostly about documentation / comments / docstrings to make everything super clear.

It's great that you already created unit tests at the `drjit-core` level!

I still have to look at the big `record_ts.*` files.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+        return this->m_internal->block_prefix_reduce(
+            vt, op, size, block_size, exclusive, reverse, in, out);
+    }
+
+    /// Compute a dot product of two equal-sized arrays
+    void reduce_dot(VarType type, const void *ptr_1, const void *ptr_2,
+                    uint32_t size, void *out) override {
+        jitc_log(
+            LogLevel::Warn,
+            "RecordThreadState::reduce_dot(): unsupported function recording!");
```
merlinND, line 451: Shouldn't unsupported operations throw an exception? (ideally with as much information as possible to help them find where the unsupported op was triggered).

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            scoped_pause();
+            this->m_internal->memcpy_async(dst, src, size);
+        }
+        if (!paused() && !has_var) {
+            // If we did not know the source variable, this memcpy might be
+            // coming from a `jitc_call_upload` call.
+            // If that is the case, we have to capture the offset buffer.
+            // Since the pointer might be used, for example by an aggregate call
+            // (nested calls), we have to overwrite the RecordVariable.
+            //
```
merlinND, line 400:
```suggestion
```

src/record_ts.h
merlinND: For unsupported operations, you could add a comment explaining why it would be difficult / impossible to support them.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+        }
+        scoped_pause();
+        return this->m_internal->mkperm(values, size, bucket_count, perm,
+                                        offsets);
+    }
+
+    /// Perform a synchronous copy operation
+    void memcpy(void *dst, const void *src, size_t size) override {
+        jitc_log(LogLevel::Debug, "record(): memcpy(dst=%p, src=%p, size=%zu)",
+                 dst, src, size);
```
merlinND, line 358: Consistently remove or keep the logs

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+        return this->m_internal->memcpy(dst, src, size);
+    }
+
+    /// Perform an asynchronous copy operation
+    void memcpy_async(void *dst, const void *src, size_t size) override {
+        jitc_log(LogLevel::Debug,
+                 "record(): memcpy_async(dst=%p, src=%p, size=%zu)", dst, src,
+                 size);
+        bool has_var = has_variable(src);
+        if (!paused() && (has_var)) {
```
merlinND, line 369:
```suggestion
        if (!paused() && has_var) {
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            for (CallData *tmp : calls_assembled) {
+                if (tmp->offset == dst) {
+                    call = tmp;
+                    break;
+                }
+            }
+            if (call) {
+                capture_call_offset(dst, size);
+                jitc_log(LogLevel::Debug, "    captured call offset");
+            }
```
merlinND, line 411: Error message if we still didn't find the variable?

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+                record_exception();
+            }
+        }
+        scoped_pause();
+        this->m_internal->aggregate(dst, agg, size);
+    }
+
+    // Enqueue a function to be run on the host once backend computation is done
+    void enqueue_host_func(void (*callback)(void *), void *payload) override {
+        scoped_pause();
```
merlinND, line 478: Should this have a warning / exception that `enqueue_host_func` isn't recorded? Or is it safe to have it not recorded and we can still replay later without issues?

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+    bool check_kernel_cache();
+};
+
+/**
+ * \brief This struct is a wrapper arround a \ref ThreadState that records the
+ * operations performed with it.
+ */
+struct RecordThreadState : ThreadState {
+
+    RecordThreadState(ThreadState *internal) {
```
merlinND, line 254: Specify that this ThreadState does not take ownership of `internal`, so the `internal` ThreadState must stay alive as long as this new one is alive.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            Variable *v          = jitc_var(input);
+            RecordVariable rv;
+            rv.state      = RecordVarState::Input;
+            rv.init       = RecordVarInit::Input;
+            rv.index      = input_index;
+            rv.type       = (VarType) v->type;
+            uint32_t slot = this->add_variable(v->data, rv);
+            jitc_log(LogLevel::Info,
+                     "record(): Adding variable %u <%p> input %u to slot s%u",
+                     input, v->data, input_index, slot);
```
merlinND, line 541: Should probably have a much lower log level

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            this->m_recording.inputs.push_back(slot);
+        } catch (...) {
+            record_exception();
+        }
+    }
+    /**
+     * Adds an output to the recording.
+     * The output can be seen as a final operation, which also has to infer the
+     * size of it's input variables.
```
merlinND, line 550:
```suggestion
     * size of its input variables.
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+    /**
+     * This captures the offset buffer of a vcall in a kernel.
+     * The offset buffer describes where in the data buffer of that vcall the
+     * variables or pointers to variables, for that vcall are stored.
+     * It should not change between invocations and we should therefore be able
+     * to capture it and reuse it when replaying the kernel.
+     */
+    uint32_t capture_call_offset(const void *ptr, size_t dsize) {
+        uint32_t size;
+        size = dsize / type_size[(uint32_t) VarType::UInt64];
```
merlinND, line 667:
```suggestion
        uint32_t size = dsize / type_size[(uint32_t) VarType::UInt64];
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+    void record_reduce_expanded(VarType vt, ReduceOp reduce_op, void *data,
+                                uint32_t exp, uint32_t size);
+
+    /**
+     * This captures the offset buffer of a vcall in a kernel.
+     * The offset buffer describes where in the data buffer of that vcall the
+     * variables or pointers to variables, for that vcall are stored.
+     * It should not change between invocations and we should therefore be able
+     * to capture it and reuse it when replaying the kernel.
+     */
```
merlinND, line 664: Specify what is `dsize` and its unit (number of bytes?)

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            slot = this->m_recording.record_variables.size();
+
+            this->m_recording.record_variables.push_back(rv);
+
+            this->ptr_to_slot.insert({ ptr, slot });
+        } else {
+            slot                = it.value();
+            RecordVariable &old = this->m_recording.record_variables[slot];
+            if (old.init != RecordVarInit::None)
+                jitc_fail("record(): Tried to overwrite a initialized variable "
```
merlinND, line 695:
```suggestion
                jitc_fail("record(): Tried to overwrite an initialized variable "
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+                              bool /*overwrite*/ = false) {
+
+        scoped_pause();
+        Variable *v = jitc_var(index);
+        if (v->scope < this->m_internal->scope && test_scope) {
+            jitc_raise(
+                "record(): Variable r%u[%u] -> %p, label=%s, was created "
+                "before recording was started, but it was "
+                "not speciefied as an input variable! This can happen if a "
+                "input type is not fully traversavle, for example when not "
```
merlinND, line 720:
```suggestion
                "not specified as an input variable! This can happen if a "
                "input type is not fully traversable, for example when not "
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+                "specifying a member in DRJIT_STRUCT, but using it in the "
+                "frozen function.",
+                index, v->size, v->data, jitc_var_label(index));
+        }
+
+        jitc_raise("record(): Variable r%u[%u] -> %p, label=%s, data=%s, of "
+                   "size > 1 was created while recording.",
+                   index, v->size, v->data, jitc_var_label(index),
+                   jitc_var_str(index));
```
merlinND, line 729: Is this something the user might encounter? It would be great to have a helpful error message like the one above.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+    }
+
+    /**
+     * Add information about a variable, deduplicating it and returning the slot
+     * in the `variables` field of the recording.
+     * Information is combined when the variable has already been added.
+     * This is used by the input variables of a kernel.
+     */
+    uint32_t add_variable(const void *ptr, RecordVariable rv) {
+
```
merlinND, line 741:
```suggestion
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            uint32_t slot = it.value();
+
+            this->m_recording.record_variables[slot] |= rv;
+
+            return slot;
+        }
+    }
+
+    /// Return the slot index given the data pointer of a variable.
+    /// This fails if the variable has not been added.
```
merlinND, line 763:
```suggestion
    /// This fails if the variable had not been previously added.
```

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+
+        } else if (info.type == ParamType::Input) {
+
+            jitc_log(LogLevel::Debug, " -> param s%u", info.slot);
+
+            if (info.test_uninit && rv.state == RecordVarState::Uninit)
+                jitc_raise("record(): Varaible at slot s%u was read from by "
+                           "operation o%u, but has not yet been initialized! "
+                           "This can happen if the variable was not part of "
+                           "the input but is used by an recorded operation.",
```
merlinND, line 806: Explain how the user could make it "part of the input", if possible.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+        add_param(info);
+    }
+    /// Helper function recording input access given the pointer.
+    void add_in_param(const void *ptr, VarType vtype = VarType::Void,
+                      bool test_uninit = true) {
+        uint32_t slot = this->get_variable(ptr);
+        add_in_param(slot, vtype, test_uninit);
+    }
+    /// Helper function recording an output access, given the slot and \ref
+    /// VarType
```
merlinND, line 833:
```suggestion
    /// Helper function recording an output access, given the slot and \ref VarType
```
and same below

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+
+bool jitc_freeze_pause(JitBackend backend);
+
+bool jitc_freeze_resume(JitBackend backend);
+
+void jitc_freeze_replay(Recording *recording, const uint32_t *inputs,
+                        uint32_t *outputs);
+
+int jitc_freeze_dry_run(Recording *recording, const uint32_t *inputs,
+                        uint32_t *outputs);
```
merlinND, line 873: Maybe move these to the top to make them more noticeable, since `RecordThreadState` is so long.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+        ReduceOp rtype;
+        struct {
+            /// The reduce type of a prefix reduction operation
+            ReduceOp rtype;
+            /// Weather a prefix sum operation is exclusive
+            bool exclusive;
+            bool reverse;
+        } prefix_reduce;
+        /// The bucket count for the mkperm operation
+        uint32_t bucket_count;
```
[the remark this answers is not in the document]
DoeringChristian: Good point. As long as it comes from python it should be fine. If it is determined by the width of some variable this might be a problem.

src/record_ts.h
DoeringChristian: Not sure, does this make sense?
```c++
                jitc_raise("record(): Varaible at slot s%u was read from by "
                           "operation o%u, but has not yet been initialized! "
                           "This can occur if the variable was not part of "
                           "the input but is used by a recorded operation, for "
                           "example if it was not specified as a member in a "
                           "DRJIT_STRUCT but used in the frozen function.",
                           info.slot,
                           (uint32_t) this->m_recording.operations.size());
```

merlinND: Okay! Fixed a couple of typos below:

```cpp
                jitc_raise("record(): Variable at slot s%u was read by "
                           "operation o%u, but it had not yet been initialized! "
                           "This can occur if the variable was not part of "
                           "the input but is used by a recorded operation, for "
                           "example if it was not specified as a member in a "
                           "DRJIT_STRUCT but used in the frozen function.",
                           info.slot,
                           (uint32_t) this->m_recording.operations.size());
```

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+        this->data_size  = 0;
+        this->alloc_size = 0;
+    }
+};
+
+/// Kernel parameter buffer
+static std::vector<void *> kernel_params;
+
+/// Temporary variables used for replaying a recording.
+static std::vector<ReplayVariable> replay_variables;
```
merlinND, line 133: What happens if someone tries replaying two different recordings from two different threads?
(I'm not sure what thread-safety the new DrJit version provides)

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                                             op.kernel.key->flags));
+                if (it == state.kernel_cache.end())
+                    jitc_raise(
+                        "replay(): Freezing functions relies on the kernel "
+                        "cache. It looks like this has been cleared since "
+                        "recording the function.");
+
+                // Reconstruct the \ref kernel_params for this launch given the
+                // allocations when replaying.
+                kernel_params.clear();
```
merlinND, line 202: Is there any risk this interferes with an unrelated pending kernel launch?

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+                if (backend == JitBackend::CUDA) {
+                    // First parameter contains kernel size.
+                    // Assigned later.
+                    kernel_params.push_back(nullptr);
+                } else {
+                    // First 3 parameters reserved for: kernel ptr, size, ITT
+                    // identifier
```
merlinND, line 210:
```suggestion
                    // First 3 parameters reserved for: kernel ptr, size, ITT identifier
```

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+    // The main loop that executes each operation
+    for (uint32_t i = 0; i < this->operations.size(); ++i) {
+        Operation &op = this->operations[i];
+        ProfilerPhase profiler(pr_operation);
+        if (!op.enabled)
+            continue;
+
+        switch (op.type) {
+            case OpType::KernelLaunch: {
```
merlinND, line 185: Move this case's handling to a function, since it's quite involved?
The same is done in e.g. `cuda_eval.cpp`, there are local functions for the more complex cases.

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                        rv.data != nullptr || dry_run,
+                        "replay(): Encountered nullptr in kernel parameters.");
+                    kernel_params.push_back(rv.data);
+                }
+
+                // Change kernel size in `kernel_params`
+                if (backend == JitBackend::CUDA) {
+                    uintptr_t size = 0;
+                    std::memcpy(&size, &launch_size, sizeof(uint32_t));
+                    kernel_params[0] = (void *) size;
```
merlinND, line 327: I'm not 100% sure I understand this. Is it essentially writing `launch_size` to `kernel_params[0]`, but masking out bits 32...63?
Wouldn't it be better to throw an exception if `launch_size >= 2**32` and directly write `kernel_params[0] = launch_size;`?

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                    }
+                    ts->launch(kernel, op.kernel.key, op.kernel.hash,
+                               launch_size, &kernel_params, &kernel_param_ids);
+                    if (op.uses_optix)
+                        uses_optix = false;
+                }
+
+            };
+            break;
+            case OpType::Barrier: {
```
merlinND, line 352: Empty line between cases (because each block is pretty large).

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                ptr_var.alloc(backend, op.size, op.input_size);
+
+                jitc_log(LogLevel::Debug,
+                         "replay(): MemsetAsync -> s%u [%zu], "
+                         "op.input_size=%zu",
+                         ptr_info.slot, op.size, op.input_size);
+
+                uint32_t size = ptr_var.size(op.input_size);
+
+                if (!dry_run)
```
merlinND, line 375: I guess the rest of the cases' code is still executed in dry_run mode so that you can see the logs?
Other than the log, it's a no-op right?

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                    ts->memset_async(dst_rv.data, new_size, tsize, &op.data);
+
+                if (!dry_run && memcpy)
+                    ts->memcpy_async(dst_rv.data, src_ptr,
+                                     (size_t) size * tsize);
+
+                dst_rv.data_size = size * type_size[(uint32_t) dst_info.vtype];
+                // dst_rv.size = size;
```
merlinND, line 460: Was this something to look into still, or should it be removed?

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+    /// same size as the number of inputs and outputs with which the frozen
+    /// function was recorded.
+    int replay(const uint32_t *replay_input, uint32_t *outputs);
+
+    /// Counter, counting the number of kernels for debugging.
+    uint32_t n_kernels = 0;
+
+    /// This function is called after recording and checks that the recording is
+    /// valid i.e. that no variables where left uninitialized.
+    void validate();
```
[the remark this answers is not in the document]
DoeringChristian: I don't think this will improve much, as the function is called once after recording to check that there are no uninitialized variables.

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+                ReplayVariable &in_rv  = replay_variables[in_info.slot];
+                ReplayVariable &out_rv = replay_variables[out_info.slot];
+
+                uint32_t size = in_rv.size(in_info.vtype);
+                out_rv.alloc(backend, size, out_info.vtype);
+
+                if (dry_run)
+                    jitc_fail(
+                        "replay(): dry_run compress operation not supported!");
```
merlinND, line 479: If the user runs into this error, what could they do about it?
I guess the error message should explain what kind of operations require a dry-run, and say that you can't combine these dry-run ops together with `compress` in the same frozen function?

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+                // size_t size = src_var.size(src_info.vtype);
+                jitc_log(LogLevel::Debug,
+                         "replay(): MemcpyAsync s%u <- s%u [%zu]",
+                         dst_info.slot, src_info.slot, src_var.data_size);
+
+                dst_var.alloc(backend, src_var.data_size);
+
+                jitc_log(LogLevel::Debug, "    src.data=%p", src_var.data);
+                jitc_log(LogLevel::Debug, "    dst.data=%p", dst_var.data);
```
merlinND, line 506: Reorganize a bit so that the 3 logs can be combined?

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                         values_info.slot, values_var.data, size);
+
+                jitc_log(LogLevel::Info, " <- perm: s%u", perm_info.slot);
+                perm_var.alloc(backend, size, perm_info.vtype);
+
+                jitc_log(LogLevel::Info, " <- offsets: s%u", offsets_info.slot);
+                offsets_var.alloc(backend, bucket_count * 4 + 1,
+                                  offsets_info.vtype);
+
+                jitc_log(LogLevel::Debug,
```
merlinND, line 541: Combine all the log calls

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+            m_exception = std::current_exception();
+    }
+
+    bool m_paused = false;
+
+    std::exception_ptr m_exception = nullptr;
+
+    ThreadState *m_internal;
+
+    Recording m_recording;
```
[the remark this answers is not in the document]
DoeringChristian: They have to be accessed by the `jitc_freeze_*` functions (except `m_pause` I forgot that).

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+                ReplayVariable &src_var = replay_variables[src_info.slot];
+                ReplayVariable &dst_var = replay_variables[dst_info.slot];
+
+                // size_t size = src_var.size(src_info.vtype);
+                jitc_log(LogLevel::Debug,
+                         "replay(): MemcpyAsync s%u <- s%u [%zu]",
+                         dst_info.slot, src_info.slot, src_var.data_size);
+
+                dst_var.alloc(backend, src_var.data_size);
```
merlinND, line 503: Could it be a problem that the memory is allocated even during a dry-run? (for this and other ops)
Will it be quickly freed again? We don't want to accidentally double memory usage because of dry runs.

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+
+                if (!dry_run)
+                    ts->block_prefix_reduce(
+                        out_info.vtype, op.prefix_reduce.rtype, size,
+                        block_size, op.prefix_reduce.exclusive,
+                        op.prefix_reduce.reverse, in_var.data, out_var.data);
+
+            };
+            break;
+            case OpType::Aggregate: {
```
merlinND, line 620: Move contents to a dedicated function since it's a complex case

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                ReplayVariable &rv = replay_variables[info.slot];
+
+                rv.free();
+
+            };
+            break;
+            default:
+                jitc_fail(
+                    "An operation has been recorded, that is not known to "
+                    "the replay functionality!");
```
merlinND, line 711: Print the op type to help with debugging

merlinND (commented):
Review part 3.
It's a partial review of `record_ts.cpp`, but unfortunately I won't have time to review the rest of the file, there's too much code.

src/record_ts.h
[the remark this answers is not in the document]
merlinND: Makes sense for `validate()`, but about the rest of the launch code, is there anything we could skip to make launches faster when a special "I know what I'm doing" flag is set by the user?

src/record_ts.h
DoeringChristian: As far as I can tell, this function is only called when freeing numpy arrays, which should be fine during recording.

tests/vcall.cpp
```diff
@@ -1251,3 +1252,118 @@ TEST_BOTH(13_load_bool_data) {
...
+        Base *base = (Base *) self;
+        UInt32 i0  = UInt32::borrow(inputs[0]);
+        UInt32 o0  = base->f(i0);
+        jit_var_inc_ref(o0.index());
+
+        outputs[0] = o0.index();
+    };
+
+    {
+        BasePtr self = arange<UInt32>(10) % 3;
```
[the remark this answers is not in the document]
DoeringChristian: Added this comment:
```cpp
        /**
         * Between recording and replay, the registry has to stay the same i.e.
         * instances should not be added or removed.
         * This is ensured by the implementation in drjit.
         * The behavior is undefined if instances are added or removed between
         * recording and replay.
         */
```

tests/vcall.cpp
```diff
@@ -1251,3 +1252,118 @@ TEST_BOTH(13_load_bool_data) {
...
+        UInt32 r0(0, 2, 4, 0, 5, 7, 0, 8, 10, 0);
+
+        jit_log(LogLevel::Info, "Recording:");
+
+        uint32_t inputs[] = {
+            self.index(),
+            i0.index(),
+        };
+
+        jit_freeze_start(Backend, inputs, 2);
```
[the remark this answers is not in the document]
DoeringChristian:
```cpp
        // This function does not perform any validation on the input.
        // The input should however be evaluated before starting freezing.
```

include/drjit-core/jit.h
```diff
@@ -2474,6 +2479,140 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
...
+ * \brief Start recording operations which can be replayed later without tracing
+ * operations.
+ *
+ * \param backend
+ *      The backend for which recording should be started.
+ *
+ * \param inputs
+ *      An array of input variable indices.
+ *      They have to be specified before starting the recording and can be
+ *      changed when replaying the recording.
```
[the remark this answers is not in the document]
DoeringChristian: Not quite sure about the formulation:
```cpp
 * \param inputs
 *      An array of input variable indices, which have to be
 *      specified when starting the recording.
 *      Different indices, representing other variables, 
 *      might be provided when replaying.
```

include/drjit-core/jit.h
```diff
@@ -2474,6 +2479,140 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
...
+ *      The size of the array has to be equal to the number of output variables
+ *      supplied to the jit_record_stop function.
+ */
+extern JIT_EXPORT void jit_freeze_replay(Recording *recording,
+                                         const uint32_t *inputs,
+                                         uint32_t *outputs);
+
+/**
+ * \brief Perform a dry run of a recording (if required), which does not perform
+ * any actual work.
```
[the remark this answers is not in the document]
DoeringChristian:
```
 * \brief Perform a dry run replay of a recording (if required), which does not
 * perform any actual work.
 *
 * Dry running the recording, calculates the widths of all kernels, and might
 * return 0 if the function has to be re-recorded.
 * This is required to catch cases where the size of an input variable changes
 * the compiled kernels, for example when performing scatter reductions in LLVM
 * mode.
```

include/drjit-core/jit.h
```diff
@@ -2474,6 +2479,140 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
...
+ *
+ * \return false if retracing the function is required, true otherwise
+ */
+extern JIT_EXPORT int jit_freeze_dry_run(Recording *recording,
+                                         const uint32_t *inputs,
+                                         uint32_t *outputs);
+
+/**
+ * \brief Pause recording the ThreadState for this backend.
+ *      Returns true if recording has already been paused.
```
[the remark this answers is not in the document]
DoeringChristian: It's possible to pause the recording twice, but the second call will not have any effect. The function returns the previous `m_pause` state of the `RecordThreadState`. I'm not sure how to formulate that in a concise way.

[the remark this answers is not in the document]
merlinND: I gave it a try:

```cpp
/**
 * \brief Pause recording the ThreadState for this backend.
 *  Returns a boolean indicating the pause state before calling
 *  this function, i.e. returns ``true`` if recording was already paused.
 * If no recording is in process, this function throws an exception.
 *
 * \param backend
 *      The backend for which to pause recording the thread state.
 */
```

(assuming it actually throws an exception)

src/record_ts.h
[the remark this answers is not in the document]
merlinND: Cool, could you please add a comment that explains it?

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+        uint32_t inputs[] = {i0.index()};
+        uint32_t outputs[1];
+
+        jit_freeze_replay(recording, inputs, outputs);
+
+        jit_log(LogLevel::Info, "o0: %s", jit_var_str(outputs[0]));
+        jit_assert(jit_var_all(jit_var_eq(r0.index(), outputs[0])));
+    }
+
+    jit_freeze_destroy(recording);
```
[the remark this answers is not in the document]
DoeringChristian: I've implemented a minimal `FrozenFunction` this allows us to implement very simple tests. However this change will make most comments on the `record.cpp` file outdated.

src/record_ts.cpp
```diff
@@ -0,0 +1,1973 @@
...
+                return false;
+        }
+    }
+    return true;
+}
+
+void jitc_freeze_start(JitBackend backend, const uint32_t *inputs,
+                       uint32_t n_inputs) {
+
+    if (jitc_flags() & (uint32_t) JitFlag::FreezingScope)
```
wjakob, line 1620: Does this apply to the parent project as well? -- i.e., we cant call a `@dr.freeze`-annotated function from another one?

wjakob (commented):
Hi Christian,

this is very impressive work! Here is a first batch of feedback, much of it about the structure and comments.

include/drjit-core/jit.h
```diff
@@ -2474,6 +2480,154 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
...
+                                         uint32_t *outputs);
+
+/**
+ * \brief Perform a dry run replay of a recording (if required), which does not
+ * perform any actual work.
+ *
+ * Dry running the recording, calculates the widths of all kernels, and might
+ * return 0 if the function has to be re-recorded.
+ * This is required to catch cases where the size of an input variable changes
+ * the compiled kernels, for example when performing scatter reductions in LLVM
```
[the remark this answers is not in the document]
DoeringChristian: That's just an example I could come up with.
Currently we only return `false` in the dry run if we replay a `Expand` operation and the size changed (this is specific to LLVM).
We also return false for the `Compress` operation, since we cannot infer the size of the resulting variable in the dry run mode and have to retrace the kernel in that case.
Dry runs are only performed if a `Expand` operation has been detected, so this will only be of concern for the `LLVM` backend.

Should I add something like this to the comment as well?

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+/// Output variables are only tracked through the outputs array, as this
+/// information is only needed when constructing the output variables.
+enum class RecordedVarState {
+    /// This variable was not initialized
+    Uninitialized,
+    /// This variable has been created by an operation
+    OpOutput,
+    /// This variable is part of the function input
+    Input,
+    /// This variable has been captured i.e. it is copied and part of the
```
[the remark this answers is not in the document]
DoeringChristian: For the `Operation` comment is it fine to leave it like this?:
```cpp
/**
 * Represents an operation that was recorded by a \ref RecordThreadState.
 *
 * To record an operation, a function in the \ref RecordThreadState has to first
 * get the current size of the depenceny vector.
 * uint32_t start = m_recording.dependencies.size();
 *
 * Then, the parameters can be added.
 * add_out_param(ptr_id, VarType::Void);
 *
 * Finally, the size after adding the parameters can be read and the operation
 * can be pushed.
 * uint32_t end = m_recording.dependencies.size();
 *
 * Operation op;
 * op.dependency_range = std::pair(start, end);
 */
```

src/record_ts.cpp
DoeringChristian: No that's handled by the feature in `drjit` it will test if we are in a `FreezingScope` already and not attempt to start another recording but just call the inner function. Should that be a comment here?

DoeringChristian: Regarding the warning during dry run when replaying compressions. This should be fine, since returning `false` in dry run mode will cause a re-trace of the function. It will however be a bit slower (it has to do the dry run) that's why I put the warning there (I'll expand on it in the warning message).

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+    bool check_kernel_cache();
+};
+
+/**
+ * \brief This struct is a wrapper arround a \ref ThreadState that records the
+ * operations performed with it.
+ */
+struct RecordThreadState : ThreadState {
+    
+    /// The last exception thrown while recording.
```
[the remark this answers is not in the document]
DoeringChristian: Expanded on the comment. 
Atm. the exception is re-thrown when ending the recording.
I'm currently propagating it to python, but we could just not safe the recording and proceed with normal evaluation/tracing, we only capture exceptions for the `record_*` functions.

include/drjit-core/jit.h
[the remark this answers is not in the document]
DoeringChristian: Added some comments regarding this (not sure if they are enough).

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+            i1.index(),
+        };
+        uint32_t outputs[2];
+
+        jit_freeze_replay(recording, inputs, outputs);
+
+        jit_log(LogLevel::Info, "o0: %s", jit_var_str(outputs[0]));
+        jit_log(LogLevel::Info, "o1: %s", jit_var_str(outputs[1]));
+        jit_assert(jit_var_all(jit_var_eq(r0.index(), outputs[0])));
+        jit_assert(jit_var_all(jit_var_eq(r1.index(), outputs[1])));
```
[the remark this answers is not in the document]
DoeringChristian: After adding the minimal `FrozenFunction` implementation, the tests look like this:

```cpp
TEST_BOTH(9_input_passthrough) {

    auto func = [](UInt32 x) {
        auto y = x + 1;
        return std::tuple(y, y, x);
    };

    FrozenFunction frozen(Backend, func);

    for (uint32_t i = 0; i < 3; i++) {
        auto x = arange<UInt32>(10 + i);
        x.make_opaque();

        auto result = frozen(x);

        auto reference = func(x);

        jit_assert(jit_var_all(jit_var_eq(std::get<0>(result).index(),
                                          std::get<0>(reference).index())));
        jit_assert(jit_var_all(jit_var_eq(std::get<1>(result).index(),
                                          std::get<1>(reference).index())));
        jit_assert(jit_var_all(jit_var_eq(std::get<2>(result).index(),
                                          std::get<2>(reference).index())));
    }
}
```
Do you think I should still add "check" functions or is it better to write the asserts manually?

tests/record.cpp
```diff
@@ -0,0 +1,553 @@
...
+        uint32_t inputs[] = {
+            i0.index(),
+        };
+
+        jit_freeze_start(Backend, inputs, 1);
+
+        UInt32 tmp = i0 + 1;
+        tmp.schedule();
+        jit_eval();
+        UInt32 o0 = tmp + 1;
```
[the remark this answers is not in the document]
DoeringChristian: Something like this?
```cpp
    auto func = [](UInt32 x) {
        auto y = x + 1;
        y.eval();
        return y + x;
    };
```

src/record_ts.cpp
DoeringChristian: Its only accessed in `jitc_*` functions, which are behind the `state.lock` mutex.

src/record_ts.cpp
DoeringChristian: That's not the same `kernel_params `as in `eval.cpp`, as I didn't want to expose that in `eval.h` it should not clear that one. Though we could make it the same vector, since we cannot be in `jitc_eval` and `jitc_freeze_replay` at the same time due to the `state.lock` mutex.

src/record_ts.cpp
DoeringChristian: I just copied what is in `eval.cpp`:
```cpp
    if (backend == JitBackend::CUDA) {
        uintptr_t size = 0;
        memcpy(&size, &group.size, sizeof(uint32_t));
        kernel_params.push_back((void *) size);

        // The first 3 variables are reserved on the CUDA backend
        n_regs = 4;
    } else {...
```
As far as I understand, the `memcpy` ensures that we copy to the first 4 bits of the first void pointer in `kernel_params` (endianess?).

src/record_ts.cpp
DoeringChristian: The rest of the `replay_*` functions might still be performing size inference, for example calling `rv.alloc(backend, size)` will not allocate in dry run mode, but will change the `dsize` of the variable. Calling `rv.size(vartype)` in the next operation will then calculate the size from the memory size and type.

Should I add a comment?

src/record_ts.cpp
DoeringChristian: This should be fine, since returning `false` in dry run mode will cause a re-trace of the function. It will however be a bit slower (it has to do the dry run) that's why I put the warning there (I'll expand on it in the warning message).

Does this make sense?
```cpp
                    jitc_log(LogLevel::Warn,
                             "replay(): Could not infer the size of the result "
                             "to a dr.compress call during dry run. The "
                             "function will be re-recorded, which causes some "
                             "overhead. This occured most likely due to a "
                             "scatter_reduce call in a LLVM kernel in the same "
                             "frozen function. To avoid this try to split the "
                             "function into two frozen functions.");

```

src/record_ts.cpp
DoeringChristian: Ah right, I'll add a comment explaining it. `rv.alloc()` does not perform allocation when in dry run mode. It just sets the size (`data_size`) of the variable in that case.

src/record_ts.cpp
```diff
@@ -0,0 +1,1973 @@
...
+    ReplayVariable(RecordedVariable &rv) {
+        this->index = rv.index;
+
+        this->init = rv.init;
+
+        if (init == RecordedVarInit::Captured) {
+            // copy the variable, so that it isn't changed
+            this->index = jitc_var_copy(this->index);
+
+            Variable *v      = jitc_var(this->index);
```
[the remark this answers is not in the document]
DoeringChristian: Should I use the `m_*` convention? Its used sometimes by Dr.Jit, but not all the time.

include/drjit-core/jit.h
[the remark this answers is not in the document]
wjakob: Yes, that would be great! Having a central list of troublesome or potentially troublesome operations somewhere is good. (The other one I remember was "Poke").

src/record_ts.cpp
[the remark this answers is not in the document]
wjakob: I usually use the `m_*` prefix for members "classes" (heap-allocated types with complex logic implemented in various methods, potentially some virtual) while skipping it for "structs" (generally stack-allocated type, mainly for storage). The precise threshold between these two is not always 100% clear.

src/record_ts.cpp
```diff
@@ -0,0 +1,1973 @@
...
+
+                ReplayVariable &data_var = replay_variables[data_info.slot];
+
+                VarType vt       = data_info.vtype;
+                ReduceOp rop     = op.rtype;
+                uint32_t size    = data_var.size(vt);
+                uint32_t tsize   = type_size[(uint32_t) vt];
+                uint32_t workers = pool_size() + 1;
+
+                uint32_t replication_per_worker =
```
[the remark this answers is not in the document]
wjakob: Good idea. Since it is only used for the LLVM backend, it could be a global function like `jitc_llvm_expand_replication_factor` or something like that.

src/record_ts.h
```diff
@@ -0,0 +1,873 @@
...
+
+    RecordVariable() {}
+
+    /**
+     * Not all information about variables might be known right away (see
+     * memcpy). When new information about the variable is available, we can add
+     * it to the already saved RecordVariable.
+     */
+    RecordVariable &operator|=(const RecordVariable &rhs) {
+        if (this->state == RecordVarState::Uninit) {
```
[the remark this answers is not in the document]
DoeringChristian: Removed the function, as it was a bit ambiguous what it does.

src/record_ts.cpp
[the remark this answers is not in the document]
DoeringChristian: Ok, I made `Recording` a class and used `m_*` prefixes, since it is only heap allocated (and I had a name conflict in `Recording::replay`).

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+extern const char *op_type_name[(int) OpType::Count];
+
+/**
+ * Represents an operation, that was recorded by a \ref RecordThreadState.
+ *
+ * To record an operation the first step is to get the current size of the
+ * dependencies vector.
+ * uint32_t start = m_recording.dependencies.size();
+ *
+ * Then, the parameters can be added.
```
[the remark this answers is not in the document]
DoeringChristian: I referenced the example at the top of the file and explain it there.

src/record_ts.h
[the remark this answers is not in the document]
DoeringChristian: This is now also part of the "gotchas" list at the top of the file.

src/record_ts.cpp
```diff
@@ -0,0 +1,1446 @@
...
+    uint32_t index;
+    RecordVarInit init;
+
+    ReplayVariable(RecordVariable &rv) {
+        this->index = rv.index;
+
+        this->init = rv.init;
+
+        if (init == RecordVarInit::Captured) {
+            // copy the variable, so that it isn't changed
```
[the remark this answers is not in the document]
DoeringChristian: Added comment, explaining that captured variables are not supposed to change between invocations.
It also explains that they are only used to capture offset buffers for now, which are not supposed to change.

src/record_ts.cpp
DoeringChristian: They are now ordered in the following way (mkperm example):
```cpp
uint32_t RecordThreadState::mkperm(...){...}
void RecordThreadState::record_mkperm(...){...}
void RecordThreadState::replay_mkperm(...){...}
```

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+     * Reductions in LLVM might be split into three operations.
+     * First the variable is expanded by its size times the number of workers +
+     * 1 Then the kernel writes into the expanded variable with some offset, and
+     * finally the variable is reduced.
+     * The expand operation allocates a new memory region and copies the old
+     * content into it.
+     * We catch this case if the input variable of a kernel has a reduce_op
+     * associated with it.
+     * The source variable is discovered by walking the operations to the last
+     * memcpy and memset, which are then disabled by this function.
```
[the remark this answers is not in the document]
DoeringChristian: I think this explains it now:
```cpp
    /**
     * Record the Expand operation, corresponding to the ``jitc_var_expand``
     * call, with which the variable has been expanded.
     *
     * Reductions in LLVM might be split into three operations. First the
     * variable is expanded by its size times the number of workers + 1 Then the
     * kernel writes into the expanded variable with some offset, and finally
     * the variable is reduced. The expand operation allocates a new memory
     * region and copies the old content into it. We catch this case if the
     * input variable of a kernel has a ``reduce_op`` associated with it. The
     * source variable is discovered by walking the operations to the last
     * memcpy and memset, which are then disabled by this function. This has to
     * be done since these operations are replaced by the expand operation.
     */
```

include/drjit-core/jit.h
```diff
@@ -2474,6 +2480,153 @@ extern JIT_EXPORT uint32_t jit_array_write(uint32_t target, uint32_t offset,
 extern JIT_EXPORT uint32_t jit_array_read(uint32_t source, uint32_t offset,
                                           uint32_t mask);
 
+// ====================================================================
+//     Core functionality to enable kernel freezing via @dr.freeze
+// ====================================================================
+
+/// Opaque data structure, storing the recording of a thread state
```
wjakob, line 2487: "[..] storing a sequence of Dr.Jit operations recorded using the API below" (The thread state itself isn't really recorded, it's more how the thread state is used)

src/record_ts.h
```diff
@@ -0,0 +1,709 @@
+/** Frozen function backend
+ *
+ * This file implements the backend for the `FrozenFunction` feature.
```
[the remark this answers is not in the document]
wjakob: A request: could you move this comment to the `record_ts.cpp` file where most of the code is located? These here are just minor forward declarations, and it makes the comment more difficult to find.

src/record_ts.h
DoeringChristian: Changed the logging a bit. Is it consistent enough now?

include/drjit-core/jit.h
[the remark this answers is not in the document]
DoeringChristian: Added the function. Is the pass by reference approach ok? Also added a comment, is that enough or should I expand more?

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+     * This function tries to capture a variable that is not known to the
+     * recording \c ThreadState.
+     * This is unsupported for now and raises an exception.
+     */
+    uint32_t capture_variable(uint32_t index, const void * /*ptr*/ = nullptr,
+                              bool /*remember*/ = true, bool test_scope = true,
+                              bool /*overwrite*/ = false);
+
+    /**
+     * Add information about a variable, deduplicating it and returning the slot
```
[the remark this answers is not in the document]
DoeringChristian: Changed the comments quite a bit. Does it make sense now?

src/record_ts.h
```diff
@@ -0,0 +1,531 @@
...
+    Uninitialized,
+    /// This variable has been created by an operation
+    OpOutput,
+    /// This variable is part of the function input
+    Input,
+    /// This variable has been captured i.e. it is copied and part of the
+    /// recording.
+    /// For example, the offset buffer of a vcall should not change between
+    /// recording and replay and can be copied.
+    /// Captured variables are immutable and copied when replaying, so that they
```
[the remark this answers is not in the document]
DoeringChristian: I think I expanded on that in the header a bit more. Not sure if it explains it consisely?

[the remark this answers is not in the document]
wjakob: Great, this is clear now. 👍

include/drjit-core/jit.h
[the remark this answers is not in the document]
wjakob: Great, the documentation in `jit.h` looks solid.

src/record_ts.h
```diff
@@ -0,0 +1,629 @@
...
+    Aggregate,
+    Free,
+    Count,
+};
+
+extern const char *op_type_name[(int) OpType::Count];
+
+/**
+ * Represents an operation that was recorded by a \ref RecordThreadState. To
+ * record and operation, we refer to the `reduce_dot` example at the top of this
```
wjakob, line 49: There isn't such an example in this file, perhaps you meant to reference the `.cpp` file?

DoeringChristian: Ah right, I changed that before moving the top comment to the `.cpp` file. I'll adjust it.

DoeringChristian:
Squashed everything into 7 commits. The first two are the same as in #106.

wjakob:
Merged! Thank you so much for this epic PR @DoeringChristian 🥳

PR #108, opened 2024-11-05
Add `scoped_reset_thread_state`

rtabbara:
Motivated by parallel `load_dict` issues observed in Mitsuba. In LLVM, `jitc_sync_thread` will call nanothread's `task_wait`. This allows other tasks to be picked up on the same thread while waiting. But sharing `ThreadState` across incomplete tasks can be problematic. For example, we may have pushed onto the thread state's mask stack.

Instead, prior to the `task_wait` call we want to perform a reset of the thread state that is restored once we're out of scope.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -464,6 +464,15 @@ void jitc_sync_thread(ThreadState *ts) {
         Task *task = jitc_task;
         if (!task)
             return;
+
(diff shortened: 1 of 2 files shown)


wjakob (approved):
Thanks!

njroussel (approved):
LGTM :rocket:

PR #109, opened 2024-11-26
Call: allow isolation per (variant, domain)

merlinND:
In the registry, the `backend` part of the key is replaced by a string-typed `variant`.

See https://github.com/mitsuba-renderer/drjit/pull/313.

### DIFF
--- a/src/registry.cpp
+++ b/src/registry.cpp
@@ -13,27 +13,23 @@
 #include "registry.h"
 #include "log.h"
 
-// Dr.Jit maintains an ID registry per variant, class and scope.
+// Dr.Jit maintains an ID registry per variant and domain (e.g. class).
 // This class separates multiple parallel data structures maintaining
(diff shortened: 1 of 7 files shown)


src/registry.cpp
```diff
@@ -7,25 +7,29 @@
...
-            return k1.backend == k2.backend && strcmp(k1.domain, k2.domain) == 0;
+            return (strcmp(k1.variant, k2.variant) == 0)
+                && (strcmp(k1.domain, k2.domain) == 0);
         }
     };
 
     struct Hash {
         size_t operator()(DomainKey k) const {
-            return hash_str(k.domain, (size_t) k.backend);
+            return hash_str(k.variant, 0) ^ hash_str(k.domain, 1);
```
merlinND, line 32: @wjakob Is this a correct way to combine two `hash_str()` results?

PR #110, opened 2024-12-05
Scatter: Avoid evaluation within a symbolic scope

rtabbara:
When performing scatters within symbolic loops, there is already an implied precondition that writes are non-conflicting across iterations.

So if a user want to call `scatter` multiple times in the loop body the same precondition should be assumed.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -2385,10 +2385,12 @@ uint32_t jitc_var_scatter(uint32_t target_, uint32_t value, uint32_t index_,
     // operation like dr.if_stmt()), then use that instead.
     target_v = jitc_var(target);
 
+    // Flush any potential conflicting writes by evaluating the target before the scatter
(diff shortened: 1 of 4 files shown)


wjakob:
Hi Rami,

thinking more about this, I think it is good to try to evaluate when possible. For example,

```
dr.scatter(x, ...)
while ....:
    dr.scatter(x, ..)
```

Here, we can do the first scatter outside of the symbolic region when getting to the second one inside. Can you modify the approach so that the scatters do not raise an error message if evaluating fails, and if we're in a symbolic scope?

rtabbara:
Thanks Wenzel, that's a good point. I hadn't considered that. I've updated the PR based on your suggestion. I was planning on adding some Dr.Jit python tests related to this but I've also added a (crude) test here to check the expected behaviour of the kernel splitting for scatters to the same target outside and within the loop body.

Issue #113, opened 2024-12-17
`scatter_inc` with all-false literal `active` mask


merlinND:
Calling `scatter_inc` with a literal all-False mask returns an array of size 0, which is unexpected.
I would have expected `scatter_inc(..., False)` to boil down to a `gather` instead.

We are running into this case in real code, and the zero-sized array causes a crash when used later in the code.
This was found by @faycalaa.

### Likely root cause

It is most likely due to this early exit:

https://github.com/mitsuba-renderer/drjit-core/blob/61570f7fa3273009b4d92cb25446d54b3448f08f/src/op.cpp#L2084-L2085

I think that something like this would make more sense (untested):

```cpp
    if (mask_v->is_literal() && mask_v->literal == 0)
        return steal(jitc_var_gather(target, index, /* true-valued literal for the gather */));
```

### Reproducer

```py
def reproduce(opaque: bool):
    counter = dr.ones(mi.UInt, 2)
    # dr.make_opaque(counter)  # No effect
    index = dr.zeros(mi.UInt, 2)

    # All-false literal mask triggers an "optimization" that is not correct.
    active = dr.full(mi.Bool, False, 2)
    if opaque:
        dr.make_opaque(active)

    old_value = dr.scatter_inc(counter, index, active)
    dr.eval(old_value, counter)

    print(f"{opaque=}:")
    print(f"- {old_value=}")
    print(f"- {counter=}")


reproduce(True)
reproduce(False)
```

Results in:

```
opaque=True:
- old_value=[0, 0]
- counter=[1, 1]
opaque=False:
- old_value=[]
- counter=[1, 1]
```

wjakob:
Scatter-inc does an optimization that collapses writes to same index. You should not expect to get usable data for entries with the mask set to false.

wjakob:
Right, that's a typo. It's meant to say "in the masked case"

wjakob:
By the way: you can easily combine this with a gather to get the behavior your want. But of course this has a perf cost.

merlinND:
Okay, that makes sense!
In this case we're not actually relying on the return values, so the current behavior is almost fine. 

It's just the particular case of an all-false literal `active` mask which leads an empty array to be returned (see first post):

```
opaque=False:
- old_value=[]
- counter=[1, 1]
```

Then we run into exceptions like `operands have incompatible sizes!` in subsequent code.
Would it make sense to return an array of arbitrary values, but with the correct width?

[...]

PR #114, opened 2024-12-18
scatter_inc: return correct width even in fully masked case

merlinND:
With a small unit test.

Closes https://github.com/mitsuba-renderer/drjit-core/issues/113.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -2085,8 +2085,16 @@ uint32_t jitc_var_scatter_inc(uint32_t *target_p, uint32_t index, uint32_t mask)
     if ((VarType) index_v->type != VarType::UInt32)
         jitc_raise("jit_var_scatter_inc(): 'index' must be an unsigned 32-bit array.");
 
-    if (mask_v->is_literal() && mask_v->literal == 0)
-        return 0;
+    if (mask_v->is_literal() && mask_v->literal == 0) {
+        // The return value is undefined (no need to gather the current
+        // values at these indices), but at least let's return an array
+        // of the correct width. We use a large value to make it clear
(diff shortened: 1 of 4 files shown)


src/op.cpp
```diff
@@ -2081,8 +2081,16 @@ uint32_t jitc_var_scatter_inc(uint32_t *target_p, uint32_t index, uint32_t mask)
...
-        return 0;
+    if (mask_v->is_literal() && mask_v->literal == 0) {
+        // The return value is undefined (no need to gether the current
+        // values at these indices), but at least let's return an array
+        // of the correct width. We use a large value to make it clear
+        // that it is an arbitrary number, not the previous value of
+        // the target array.
+        return jitc_var_new_node_0(var_info.backend, VarKind::Literal,
+                                   VarType::UInt32, var_info.size, var_info.symbolic,
+                                   /* payload */ (uint32_t) -1);
```
merlinND, line 2092: Should this node depend on some of the inputs? Maybe on the `index` and `mask` at least?

njroussel: No, let's keep it as a single literal. Also, let's use `jitc_var_literal` here
```cpp
uint64_t minus_one = (uint64_t) -1;
return jitc_var_literal(var_info.backend, VarType::UInt32, &minus_one, var_info.size, 0);
```

njroussel (approved):
A couple of tiny changes, then we're good to merge this :+1:

merlinND:
Thanks @njroussel!
I've applied your fixes and the CI is passing.

njroussel:
Perfect, thanks :rocket:

PR #118, opened 2025-01-30
Fixed OptiX kernel hashing bug

DoeringChristian:
Commit [0c9c54e](https://github.com/mitsuba-renderer/drjit-core/commit/0c9c54ec5c2963dd576c5a16d10fb2d63d67166f) accidentally introduced a bug, that caused the precomputed hash of a `KernelKey` to no longer be equivalent to the actual hash of that kernel key.
This PR adds the upper 64 bits of the kernel hash to the `KernelKey` struct, memoizing the computation of the hash of the kernel string. The actual hash function is also modified to use this stored hash instead of re-computing the hash of the string.

### DIFF
--- a/src/eval.cpp
+++ b/src/eval.cpp
@@ -500,10 +500,9 @@ Task *jitc_run(ThreadState *ts, ScheduledGroup group) {
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


wjakob:
Thank you @DoeringChristian!!!

PR #115, opened 2024-12-19
Frozen ThreadState Fixes

DoeringChristian:
This PR adds various fixes to the `Frozen ThreadState` #107 PR, that appeared after merging.

- Adds logging function for `capture_call_offset` function
- Fix memory leak in `vcall` tests by freeing the copied OptiX SBT and changing `borrow` to `steal` in test
- Adds the ability to traverse the whole registry without a variant string

PR #121, opened 2025-02-03
Frozen Functions Fixes

DoeringChristian:
This PR adds various fixes to the `Frozen ThreadState` #107 PR, that appeared after merging.

- Adds logging function for `capture_call_offset` function
- Fix memory leak in `vcall` tests by freeing the copied OptiX SBT and changing `borrow` to `steal` in test
- Adds the ability to traverse the whole registry without a variant string

This PR replaces #115, and changes the source branch.


njroussel (approved):

njroussel:
Looks good to me !

We're in the middle of a release, we'll wait for it to be done before merging this.

PR #123, opened 2025-02-13
Added notify_expand to track all expand cases when recording a thread…

DoeringChristian:
This PR fixes a bug with the `RecordThreadState` frozen function feature.
When `jitc_var_expand` is used on LLVM to scatter-reduce on an array, and the pointer to that array is  not directly passed to the kernel through it's arguments, but through the data buffer generated from an `aggregate` call, it was not possible to determine that this variable should be expanded when recording the ThreadState.
Therefore, a `notify_expand` function is added to the ThreadState, that is called at the end of `jitc_var_expand` to notify the RecordThreadState of this.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -762,4 +762,4 @@ void ThreadState::reset_state() {
 #endif
 }
 void ThreadState::notify_free(const void *) { }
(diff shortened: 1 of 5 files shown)


wjakob:
Awesome, this ended up being a nice simplification as well!

PR #124, opened 2025-02-14
Fix, recording the wrong size for `expand`

DoeringChristian:
Just a one liner to fix the wrong size being recorded for jitc_var_expand calls.

### DIFF
--- a/src/cuda_tex.cpp
+++ b/src/cuda_tex.cpp
@@ -76,6 +76,20 @@ struct DrJitCudaTexture {
 
         return --n_referenced_textures > 0;
     }
+
+    /**
+     * \brief Returns the ID of a JIT variable, which encodes the address of the
+     * i-th texture object.
+     *
+     * This function returns a JIT variable ID of a literal representing the
+     * texture address. This feature is internally used to make textures
+     * compatible with frozen function recording. This function returns an
+     * owning reference, which the caller must release eventually.
+     */
+    uint32_t get_jit_pointer(uint32_t i) {
+        return jitc_var_pointer(JitBackend::CUDA, jitc_var(indices[i])->data,
+                                indices[i], false);
+    }
 };
 
 struct TextureReleasePayload {
@@ -182,7 +196,7 @@ void *jitc_cuda_tex_create(size_t ndim, const size_t *shape, size_t n_channels,
         texture->arrays[tex] = array;
 
         if (tex_channels == 1)
-            view_desc.format = format == 0 ? 
+            view_desc.format = format == 0 ?
                 CU_RES_VIEW_FORMAT_FLOAT_1X32 : CU_RES_VIEW_FORMAT_FLOAT_1X16;
         else if (tex_channels == 2)
             view_desc.format = format == 0 ?
@@ -194,8 +208,9 @@ void *jitc_cuda_tex_create(size_t ndim, const size_t *shape, size_t n_channels,
         cuda_check(cuTexObjectCreate(&(texture->textures[tex]), &res_desc,
                                      &tex_desc, &view_desc));
 
-        texture->indices[tex] = jitc_var_pointer(
-            JitBackend::CUDA, (void *) texture->textures[tex], 0, 0);
+        texture->indices[tex] =
+            jitc_var_mem_map(JitBackend::CUDA, VarType::UInt64,
+                             (void *) texture->textures[tex], 1, false);
 
         TextureReleasePayload *payload_ptr =
             new TextureReleasePayload({ texture, tex });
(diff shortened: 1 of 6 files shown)


wjakob:
This commit is marked as draft, and I think it contains some commented/unrelated change. Shall I just merge the first commit [1effdc2](https://github.com/mitsuba-renderer/drjit-core/pull/124/commits/1effdc2984fa9e55bf4be2a0fd52d89b2956e599)?

wjakob:
ping @DoeringChristian  regarding the message above

DoeringChristian:
Oh sorry, missed that. I think it makes sense to merge the first one. I kept it a draft, to collect all potential fixes that might still come up with frozen functions.

wjakob:
OK 👍. I cherry-picked the first commit from this PR into master. I amended the commit metadata to fix a typo in the title (`bing` -> `being`)

src/cuda_tex.cpp
```diff
@@ -76,6 +76,19 @@ struct DrJitCudaTexture {
 
         return --n_referenced_textures > 0;
     }
+
+    /**
+     * \brief Returns a JIT pointer to the ith texture object.
+     *
+     * We map the texture to a JIT variable, which can then be revered to by a
```
wjakob, line 83: Edited text:
```
\brief Returns the ID of a JIT variable, which encodes the address of the i-th texture object.

This function returns a JIT variable ID of a literal representing the texture address. This feature is internally used to make textures compatible with frozen function recording. This function returns an owning reference, which the caller must release eventually.
```

src/cuda_tex.cpp
```diff
@@ -249,6 +265,17 @@ void jitc_cuda_tex_get_shape(size_t ndim, const void *texture_handle,
     shape[ndim] = texture.n_channels;
 }
 
+void jitc_cuda_tex_get_indices(const void *texture_handle, uint32_t *indices) {
+    if (!texture_handle)
+        return;
+
+    DrJitCudaTexture &texture = *((DrJitCudaTexture *) texture_handle);
+
+    for (uint32_t i = 0; i < texture.n_textures; i++) {
```
wjakob, line 274: Can omit braces here.

src/record_ts.cpp
```diff
@@ -2126,6 +2061,222 @@ void RecordThreadState::add_out_param(uint32_t slot, uint32_t vtype) {
     add_out_param(slot, (VarType) vtype);
 }
 
+/**
+ * \brief A simple wrapper around a ThreadState, that disables all its
+ *     operations. This is used to prevent operations to one ThreadState being
+ *     executed while a different thread state is recorded.
+ */
+struct DisabledThreadState : ThreadState {
```
wjakob, line 2069: This is a nice idea!

src/record_ts.cpp
```diff
@@ -2126,6 +2061,222 @@ void RecordThreadState::add_out_param(uint32_t slot, uint32_t vtype) {
...
+    void notify_free(const void * /*ptr*/) override {};
+};
+
+void disable_thread_state(ThreadState **ts, JitBackend recording_backend) {
+    if (!*ts)
+        return;
+    *ts = new DisabledThreadState(*ts, recording_backend);
+}
+
+void enable_thread_state(ThreadState **ts) {
```
wjakob, line 2197: The naming of these functions is a bit weird because they exist to disable functionality instead of enabling it. Would a more logical names of these function be something like ``set_disabled_thread_state`` and ``unset_disabled_thread_state`?

src/cuda_tex.cpp
```diff
@@ -194,8 +207,11 @@ void *jitc_cuda_tex_create(size_t ndim, const size_t *shape, size_t n_channels,
         cuda_check(cuTexObjectCreate(&(texture->textures[tex]), &res_desc,
                                      &tex_desc, &view_desc));
 
-        texture->indices[tex] = jitc_var_pointer(
-            JitBackend::CUDA, (void *) texture->textures[tex], 0, 0);
+        // texture->indices[tex] = jitc_var_pointer(
```
wjakob, line 210: Note the comment here. I am also wondering about the reason for encoding the texture as an UInt64 rather than as a pointer. This could have implications on compilation of textures that would breaks the kernel caching feature. 

Explanation: Both Pointer and UInt64 variables encode integer literals. The difference is that UInt64s are baked into the generated code, while Pointer variables become inputs of the kernel. When we use textures on the GPU, we don't want to specify the memory address of the texture in the generated kernel since it won't be consistent across runs.

wjakob (commented):
Hi Christian, here is some feedback on this PR.

src/cuda_tex.cpp
DoeringChristian: We still use a pointer, the same way as before. The `get_jit_pointer` function returns a pointer to the mapped memory, and is used wherever the pointer would be used previously. For recording, the pointer has to point to a jit variable.

PR #129, opened 2025-03-04
Fixed segfault in jitc_array_read

DoeringChristian:
This PR fixes a segfault in the `jitc_array_read` function, where the `variables` vector might be reallocated when calling `jitc_var_mask_apply`. `vo` and `vs` then no longer point to valid memory.


wjakob:
Oops, good catch. Thank you for tracking this one down!

By the way, these kinds of bugs become easy to track down when turning on the `DRJIT_SANITIZE_*SAN` in combination with `DRJIT_SANITIZE_INTENSE`. That invalidates the variable data structure after basically every operaiton.

PR #131, opened 2025-03-07
Fix read-after-free in `jitc_array_write()`

merlinND:
Follow-up to https://github.com/mitsuba-renderer/drjit-core/pull/129.
The `jitc_array_write()` function was susceptible to the same issue.

This was caught by running some Mitsuba code with `asan` + DrJit's intense sanitization options.


wjakob:
Thanks!

PR #132, opened 2025-03-17
Added a new `VarType::BaseFloat` for Python type promotion

wjakob:
This commit adds a dummy member `VarType::BaseFloat` to the VarType enumeration. This type cannot be used for actual computation.

It is a placeholder to be used in situations where we haven't quite decided what floating point precision to assign to a floating point variable (E.g. a Python scalar float).

This allows us to compute things like `Float16(1.0) + 2.0` without the `2.0` being interpreted as `Float32`, causing an unwanted cast to a higher precision.

### DIFF
--- a/tests/basics.cpp
+++ b/tests/basics.cpp
@@ -505,7 +505,7 @@ TEST_BOTH_FLOAT_AGNOSTIC(06_cast) {
     };
     const char *type_names[(int) VarType::Count]{
         "Void",   "Bool",  "Int8",   "UInt8",   "Int16",   "UInt16",  "Int32",
-        "UInt32", "Int64", "UInt64", "Pointer", "", "Float16", "Float32", "Float64"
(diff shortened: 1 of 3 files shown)


tests/basics.cpp
```diff
@@ -505,11 +505,11 @@ TEST_BOTH_FLOAT_AGNOSTIC(06_cast) {
     };
     const char *type_names[(int) VarType::Count]{
         "Void",   "Bool",  "Int8",   "UInt8",   "Int16",   "UInt16",  "Int32",
-        "UInt32", "Int64", "UInt64", "Pointer", "Float16", "Float32", "Float64"
+        "UInt32", "Int64", "UInt64", "Pointer", "", "Float16", "Float32", "Float64"
```
merlinND, line 508: This could be a bit confusing if the type ever gets printed in error messages or when debugging; is there a disadvantage to giving it a readable name like `"(BaseFloat)"`?

wjakob: Fixed, thanks!

PR #133, opened 2025-03-20
Opaque Width

DoeringChristian:
This PR adds a new function `jit_var_opaque_width`, that constructs an opaque variable, corresponding to the width of another variable. It also notifies the ThreadState, to record and replay this operation.

### DIFF
--- a/include/drjit-core/array.h
+++ b/include/drjit-core/array.h
@@ -253,6 +253,11 @@ template <JitBackend Backend_, typename Value_> struct JitArray {
         return jit_var_size(m_index);
     }
 
+    auto opaque_width_() {
+        using UInt32 = JitArray<Backend_, uint32_t>;
+        return UInt32::steal(jit_var_opaque_width(m_index));
+    }
+
     void resize(size_t size) {
         uint32_t index = jit_var_resize(m_index, size);
         jit_var_dec_ref(m_index);
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1130,6 +1130,18 @@ JIT_INLINE void jit_var_dec_ref(uint32_t index) JIT_NOEXCEPT {
 #define jit_var_inc_ref jit_var_inc_ref_impl
 #endif
 
+/**
+ * \brief Lock the recursive state mutex.
+ * This can improve performance when a single thread performs a large number of
+ * Dr.Jit operations in sequence.
+ */
+extern JIT_EXPORT void jit_state_lock();
+/**
+ * \brief Unlock the recursive state mutex.
+ * This should never be called from a thread that has not locked the mutex.
+ */
+extern JIT_EXPORT void jit_state_unlock();
+
 /// Query the a variable's reference count (used by the test suite)
 extern JIT_EXPORT uint32_t jit_var_ref(uint32_t index);
 
@@ -1187,6 +1199,9 @@ extern JIT_EXPORT uint32_t jit_var_data(uint32_t index, void **ptr_out);
 
 /// Query the size of a given variable
 extern JIT_EXPORT size_t jit_var_size(uint32_t index);
+
+/// Query the size of a given variable, as an opaque variable.
+extern JIT_EXPORT uint32_t jit_var_opaque_width(uint32_t index);
 
 /// Query the type of a given variable
 extern JIT_EXPORT JIT_ENUM VarType jit_var_type(uint32_t index);
@@ -1632,9 +1647,9 @@ enum JitFlag {
     JitFlagForbidSynchronization = 1 << 17,
     JitFlagScatterReduceLocal = 1 << 18,
     JitFlagSymbolic = 1 << 19,
-    KernelFreezing = 1 << 20,
-    FreezingScope = 1 << 21,
-    EnableObjectTraversal = 1 << 22,
+    JitFlagKernelFreezing = 1 << 20,
+    JitFlagFreezingScope = 1 << 21,
+    JitFlagEnableObjectTraversal = 1 << 22,
     JitFlagShaderExecutionReordering = 1 << 23
 };
 #endif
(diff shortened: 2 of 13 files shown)


include/drjit-core/array.h
```diff
@@ -253,6 +253,11 @@ template <JitBackend Backend_, typename Value_> struct JitArray {
         return jit_var_size(m_index);
     }
 
+    auto opaque_width() {
```
wjakob, line 256: I think this is the kind of function that would end with an underscore, i.e., it's considered internal.

include/drjit-core/jit.h
```diff
@@ -1176,6 +1176,10 @@ extern JIT_EXPORT uint32_t jit_var_data(uint32_t index, void **ptr_out);
 /// Query the size of a given variable
 extern JIT_EXPORT size_t jit_var_size(uint32_t index);
 
+/// Query the size of a given variable, as an opaque variable. This also
```
wjakob, line 1179: I think you can leave the ThreadState notification comment out in `jit.h`, that is true for many other functions as well and we don't list it exhaustively.

src/api.cpp
```diff
@@ -610,6 +610,27 @@ size_t jit_var_size(uint32_t index) {
...
+    Variable *var = jitc_var(index);
+    uint32_t var_size = var->size;
+
+    // The variable has to be evaluated, to notify the ThreadState
+    jitc_var_eval(index);
+
+    uint32_t width_index = jitc_var_literal(
+        (JitBackend) var->backend, VarType::UInt32, &var_size, 1, true);
+
+    ThreadState *ts = thread_state(var->backend);
```
wjakob, line 628: Unsafe to access `var->backend` here, you must save it before.

src/record_ts.cpp
```diff
@@ -471,6 +475,41 @@ void RecordThreadState::barrier() {
     return m_internal->barrier();
 }
 
+void RecordThreadState::notify_opaque_width(uint32_t index,
```
wjakob, line 478: minor: indentation of second line

src/record_ts.cpp
```diff
@@ -471,6 +475,41 @@ void RecordThreadState::barrier() {
...
+        uint32_t end = m_recording.dependencies.size();
+
+        Operation op;
+        op.type             = OpType::OpaqueWidth;
+        op.dependency_range = std::pair(start, end);
+        m_recording.operations.push_back(op);
+    }
+}
+
+int Recording::replay_opaque_width(Operation &op){
```
wjakob, line 495: Very minor: Missing space between `){`, extra newline after start of the function

src/record_ts.h
```diff
@@ -428,6 +431,8 @@ struct RecordThreadState : ThreadState {
     void reduce_expanded(VarType vt, ReduceOp reduce_op, void *data,
                          uint32_t exp, uint32_t size) override;
 
+    void notify_opaque_width(uint32_t index, uint32_t width_index) override;
```
wjakob, line 434: Maybe a brief line comment, e.g.:

``Some kernels use the width of an array in a computation. When using the kernel freezing feature, this requires special precautions to ensure that the resulting capture remains usable with different array sizes. This notification function exists so that this special-case handling can be realized.``

Likely this should go into the top level ThreadState header declaration.

wjakob (commented):
Some comments on this PR (In addition, I left one before the review, see above)

include/drjit-core/array.h
DoeringChristian: I'm not sure I would consider it internal. It is for example used in the `record.cpp` tests, and called from an external function. But, we could of course add the underscore if you think that is better.

wjakob: I think the main question is if we expect users or Mitsuba to call this. If it is just used in tests, and Dr.Jit-internal code, then a trailing underscore is better.

include/drjit-core/jit.h
```diff
@@ -1130,6 +1130,16 @@ JIT_INLINE void jit_var_dec_ref(uint32_t index) JIT_NOEXCEPT {
 #define jit_var_inc_ref jit_var_inc_ref_impl
 #endif
 
+/**
+ * \brief Lock the recursive state mutex.
```
wjakob, line 1134: Add a comment: This can improve performance when a single thread performs a large number of Dr.Jit operations in sequence.

include/drjit-core/jit.h
```diff
@@ -1130,6 +1130,16 @@ JIT_INLINE void jit_var_dec_ref(uint32_t index) JIT_NOEXCEPT {
...
+ */
+extern JIT_EXPORT void jit_state_lock();
+/**
+ * \brief Unlock the recursive state mutex.
+ *     This should never be called from a thread that has not locked the mutex.
```
wjakob, line 1139: Remove whitespace before "This"

src/var.cpp
```diff
@@ -1393,6 +1402,26 @@ VarState jitc_var_state(uint32_t index) {
         return VarState::Unevaluated;
 }
 
+VarInfo jitc_var_info(uint32_t index) {
+    VarInfo info;
```
wjakob, line 1406: `VarInfo{}`? Otherwise this may return undefined output for some fields like `info.literal` or `info.data`

src/var.cpp
```diff
@@ -1393,6 +1402,26 @@ VarState jitc_var_state(uint32_t index) {
...
+    Variable *var = jitc_var(index);
+
+    info.backend = (JitBackend)var->backend;
+    info.type = (VarType)var->type;
+    info.state = jitc_var_state(index);
+    info.size = var->size;
+    info.is_array = var->is_array();
+    info.is_coop_vec = var->coop_vec;
+    info.unaligned = var->unaligned;
+    if(info.state == VarState::Literal)
```
wjakob, line 1417: space after `if `

wjakob (approved):
OK to merge after incorporating a few more small changes.

PR #134, opened 2025-03-23
CUDA Scatter Reduce NoConflicts - address missing fmt argument

lnuic:
Fixes an issue in the scatter_reduce implementation when using ReduceMode.NoConflicts. The problem was caused by a missing argument in the call to the formatting function (fmt).

Reproducer:
```python
import drjit as dr
from drjit.auto import Int32, UInt32
dr.set_flag(dr.JitFlag.PrintIR, True)

a = Int32([-3, -62, 5])

dr.scatter_add(target=a,
               value=Int32([0, 1, 2]), 
               index=dr.arange(UInt32, 3), 
               mode=dr.ReduceMode.NoConflicts)

print(a)
```

### DIFF
--- a/src/cuda_scatter.cpp
+++ b/src/cuda_scatter.cpp
@@ -295,7 +295,7 @@ void jitc_cuda_render_scatter_reduce(const Variable *v,
             "        $s.$s %tmp, %tmp, $v;\n"
             "        st.global.$s [%rd3], %tmp;\n"
             "    }\n",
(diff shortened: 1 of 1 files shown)


njroussel (changes requested):
I was writing a test for this in `drjit`. It actually doesn't correctly support FP16 either, I've added the necessary changes as proposals.

I'll push my `drjit` test once this is merged.

Thanks for finding this :)

PR #135, opened 2025-04-10
Add missing include when dynamic CUDA is disabled

dvicini:
When compiling without dynamic CUDA, we need to include `cudaProfiler.h` for the profiler API to be available.


wjakob:
Thanks!

PR #137, opened 2025-04-11
Update OptiX API in triangle test to 8.0

dvicini:
The optix stubs used by the `triangle.cpp` unit tests are outdated and miss a few fields that are used by optix 8.0.

This PR adds the missing fields, replicating the changes already made in Mitsuba: https://github.com/mitsuba-renderer/mitsuba3/pull/1480

Without those changes, the `triangle.cpp` test fails when run under certain sanitizers.


wjakob:
Thanks!

PR #138, opened 2025-04-11
Make sure structs are fully zero initialized

dvicini:
I learned today that C++ value initialization of a struct will only initialize the first union member.

E.g., the following code does not guarantee that all union members are zero:

```cpp
struct A {
...
}
struct B {
...
}

struct C {
  union {
    A a;
    B b;
  }
}
...

C c {};
// --> c.b might not be 0 entirely
```

This causes issues when relying on it to zero-initialize complex OptiX API structs (in particular, those with pointer members). I am running Mitsuba & Dr.Jit with a sanitizer that appears to make sure to fill any non-initialized memory with a sequence of non-zero bits to surface those issues. This makes code that relies on this initialization segfault.

This PR fixes these issues in `triangle.cpp`, I will follow up with a PR to fix the same problem in Mitsuba

### DIFF
--- a/tests/triangle.cpp
+++ b/tests/triangle.cpp
@@ -54,7 +54,8 @@ void demo() {
     // =====================================================
 
     const uint32_t triangle_input_flags[1] = { OPTIX_GEOMETRY_FLAG_DISABLE_ANYHIT };
-    OptixBuildInput triangle_input {};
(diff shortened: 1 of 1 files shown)


wjakob:
Thanks!

PR #140, opened 2025-04-16
Make float32 <-> float16 conversions public

dvicini:
This PR makes `float32_to_float16` and `float16_to_float32` public. This is because they are referred to by Mitsuba's struct converter in some cases: https://github.com/mitsuba-renderer/mitsuba3/blob/master/src/core/struct.cpp#L752

Without these changes, the struct converter cannot be compiled on older instruction sets

### DIFF
--- a/include/drjit-core/half.h
+++ b/include/drjit-core/half.h
@@ -54,17 +54,6 @@ struct half {
 
     struct init_tag {};
     static constexpr half from_binary(uint16_t value) { return { value, init_tag{} }; }
-private:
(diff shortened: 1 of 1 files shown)


wjakob:
Do you need a `public:` after moving the declaration? Otherwise this makes `friend` functions private.

dvicini:
What do you mean? All I did was move those two functions above the `private:` that was already there.

wjakob:
You are right! :)

PR #141, opened 2025-04-18
Cooperative vector API

wjakob:
This PR adds cooperative vector code generation support to Dr.Jit-Core. The main documentation and test suite of this feature is in the Dr.Jit parent project.

### DIFF
--- a/src/coop_vec.cpp
+++ b/src/coop_vec.cpp
@@ -1,0 +1,730 @@
+/*
+    src/coop_vec.cpp -- Backend-independent parts of the Cooperative Vector API
+
+    Copyright (c) 2025 Wenzel Jakob <wenzel.jakob@epfl.ch>
+
+    All rights reserved. Use of this source code is governed by a BSD-style
+    license that can be found in the LICENSE file.
+*/
+
+#include "var.h"
+#include "coop_vec.h"
+#include "internal.h"
+#include "log.h"
+#include "op.h"
+#include "optix_api.h"
+#include "optix.h"
+#include "cuda.h"
+#include <drjit-core/nanostl.h>
+
+bool jitc_coop_vec_supported(JitBackend backend) {
+    if (backend == JitBackend::CUDA)
+        return (jitc_cuda_version_major == 12 &&
+                jitc_cuda_version_minor >= 8) ||
+               jitc_cuda_version_major > 12;
+    return true;
+}
+
+// Strip away loop phi nodes to recover the nested memory buffer. This is needed
+// when cooperative vectors in symbolic loops/conditionals reference the buffer
+// backing the matrix data, which has been converted into a symbolic variable.
+static uint32_t unwrap(uint32_t index) {
+    while (true) {
+        const Variable *v = jitc_var(index);
+        if (v->kind != (uint32_t) VarKind::LoopPhi)
+            return index;
+        index = borrow(v->dep[3]);
+    }
+}
+
+uint32_t jitc_coop_vec_pack(uint32_t n, const uint32_t *in) {
+    if (n == 0)
+        jitc_raise("jit_coop_vec_pack(): vector cannot be empty!");
+    if (n > 0xFFFF)
+        jitc_raise("jit_coop_vec_pack(): cooperative vector is too large!");
+
+    for (uint32_t i = 0; i < n; ++i) {
+        if (in[i] == 0)
+            jitc_raise("jit_coop_vec_pack(): argument %u is uninitialized!", i);
+    }
+
+    const Variable *arg_v = jitc_var(in[0]);
+    if (!jitc_coop_vec_supported((JitBackend) arg_v->backend))
+        jitc_raise("jit_coop_vec_pack(): The use of cooperative vectors on "
+                   "the CUDA/OptiX backend requires CUDA 12.8 or newer "
+                   "(which corresponds to driver version R570+).");
+
+    Variable v;
+    v.kind = (uint32_t) VarKind::CoopVecPack;
+    v.type = arg_v->type;
+    v.size = arg_v->size;
+    v.backend = arg_v->backend;
+    v.array_length = n;
+    v.coop_vec = true;
+    v.optix = v.backend == (uint32_t) JitBackend::CUDA;
+
+    drjit::unique_ptr<CoopVecPackData> cvid = new CoopVecPackData();
+    bool is_literal = true;
+    uint64_t literal = arg_v->literal;
+    cvid->indices.reserve(n);
+
+    jitc_log(Debug, "jit_coop_vec_pack(): building a cooperative vector with %u elements", n);
+    for (uint32_t i = 0; i < n; ++i) {
+        uint32_t index = in[i];
+        const Variable *v2 = jitc_var(index);
+        v.size = std::max(v.size, v2->size);
+        if (v2->backend != v.backend || v2->type != v.type)
+            jitc_raise("jit_coop_vec_pack(): inputs must have compatible types and backends!");
+
+        if (!v2->is_literal() || v2->literal != literal)
+            is_literal = false;
+
+        jitc_var_inc_ref(index);
+        cvid->indices.push_back(index);
+        jitc_log(Debug, "  - entry %u: r%u", i, index);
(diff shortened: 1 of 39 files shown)


wjakob:
@merlinND I forgot to create a separate PR for this part, done now.

src/coop_vec.cpp
```diff
@@ -0,0 +1,724 @@
...
+            target = steal(jitc_var_copy(target));
+
+        if ((VarType) target_v->type != vt)
+            jitc_raise("jit_coop_vec_outer_product_accum(): source/target "
+                       "buffers have an incompatible type!");
+    }
+
+    void *p = nullptr;
+    Ref tmp = steal(jitc_var_data(target, false, &p));
+    Ref target_ptr = steal(jitc_var_pointer(backend, p, tmp, 1));
```
merlinND, line 689: Does the current implementation support multiple `outer_product_accumulate` to the same matrix?
It's quite common to evaluate the same MLP multiple times in the same forward pass (e.g. taking finite difference gradients of a neural field), so in the backward pass there would be multiple outer products targeting the same matrix that we want to keep in the same kernel.

merlinND (commented):
Review part 1 (missing `llvm/cuda_coop_vec.cpp`).

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+}
+
+void jitc_optix_render_coop_vec_accum(const Variable *v, const Variable *target,
+                                      const Variable *value, const Variable *mask) {
+    bool use_mask = !mask->is_literal() || mask->literal != 1;
+    uint32_t length = value->array_length,
+             reg_count = get_reg_count(length);
+
+    if (use_mask)
+        fmt("    @!$v bra l_$u_done;\n", mask, v->reg_index);
```
merlinND, line 133: Would this op and `outer_product_accum` benefit from more coherence within the warp? (I would expect so at least for `outer_product_accum`).
If that's the case, would it make sense to apply the mask by zero-ing out the operand so that this lane is a no-op, instead of jumping over the code? Even if OptiX can maybe wake this thread back up, presumably it's faster to let it participate directly.

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+void jitc_optix_render_coop_vec(const Variable *v, const Variable *a0,
+                                const Variable *a1, const Variable *a2,
+                                const Variable *a3) {
+    uint32_t tsize = type_size[v->type],
+             length = v->array_length,
+             op_length = length;
+
+    if ((VarKind) v->kind == VarKind::CoopVecMatVec)
+        op_length = std::max(op_length, (uint32_t) a1->array_length);
```
merlinND, line 225: Explain what kind of broadcasting this allows?

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+void jitc_optix_render_coop_vec_unpack(const Variable *v, const Variable *a0) {
+    uint32_t tsize = type_size[v->type];
+
+    put("    // coop_vec_unpack\n");
+    if (tsize == 4) {
+        fmt("    mov.b32 $v, %cv$u_$u;\n", v, a0->reg_index, (uint32_t) v->literal);
+    } else {
+        fmt("    {\n"
+            "        .reg.$b %temp;\n"
+            "        cvt.u$u.u32 %temp, %cv$u_$u;\n"
```
merlinND, line 117: In theory we could back e.g. two `f16` entries of a cooperative vector into one of these `.u32` registers, but I imagine there would be hardly any gains and it wouldn't be worth the complexity?
Just checking because registers are at a premium.

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+                put("        call (), _optix_vector_op3_ptr, (%op, %type, %size, %type, %size, %src_0, %src_1, %src_2, %dst);\n");
+                copy_from_buffer("dst", v);
+            }
+            break;
+
+        case VarKind::Bitcast:
+                for (uint32_t i =  0; i < reg_count; ++i)
+                    fmt("        mov.b32 %cv$u_$u, %cv$u_$u;\n",
+                        v->reg_index, i, a0->reg_index, i);
+            break;
```
merlinND, line 369: Isn't this already handled in `cuda_eval.cpp`? Is it a different op than the existing `jitc_var_cast()`?

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+static void copy_from_buffer(const char *name, const Variable *v) {
+    uint32_t tsize  = type_size[v->type],
+             length = v->array_length;
+
+    for (uint32_t i = 0; i < length; ++i)
+        fmt("        ld.local.$b %cv$u_$u, [$s+$u];\n", v, v->reg_index, i, name, tsize * i);
+}
+
+static void put_elem(const Variable *v, uint32_t reg_count, bool trailing_comma, bool is_return_value = false) {
```
merlinND, line 97:
```suggestion
static void put_elems(const Variable *v, uint32_t reg_count, bool trailing_comma, bool is_return_value = false) {
```

src/optix_coop_vec.cpp
wjakob: It's a bit-cast for coop vectors, which are distributed over multiple registers. The code that needs to be generated in this case is different from what is in `cuda_eval.cpp` (all OptiX coopvec-related code generation is in `optix_coop_vec.cpp` and not in `cuda_eval.cpp`).

JIT variables use the `jit_var_*` API. Cooperative vectors use the `jit_coop_vec_*` API. I added a clarification to `jit.h`.

src/optix_coop_vec.cpp
wjakob: The backend (OptiX) may or may not perform such an optimization for us. If you look at how user code interfaces with OptiX through the builtin intrinsics, this is the way to do it officially. None of the OptiX SDK examples use the `float16x2` packed arithmetic types.

src/coop_vec.cpp
wjakob: Yes, that is perfectly legal. Is there something in this code fragment that made you think otherwise?

src/optix_coop_vec.cpp
```diff
@@ -0,0 +1,488 @@
...
+            put("%u, ");
+    }
+    if (!trailing_comma)
+        buffer.rewind(2);
+}
+
+void jitc_optix_render_coop_vec_unpack(const Variable *v, const Variable *a0) {
+    uint32_t tsize = type_size[v->type];
+
+    put("    // coop_vec_unpack\n");
```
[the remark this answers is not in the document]
wjakob: Indeed. The compiler could in principle re-pack things, but such optimizations are not exposed at the OptiX intrinsic level. I will add a clarification.

src/optix_coop_vec.cpp
wjakob: This is actually not related to broadcasting. I added this comment:

``` 
// For matrix multiplications, the size of the required intrinsic (e.g., 16
// vs 64 vs N) is related to the maximum cooperative vector size of the
// input and output.
```

src/coop_vec.cpp
merlinND: I was just thinking that one `outer_product_accumulate()` op may count as a side effect to `target`. So then taking `jitc_var_data()` of pointer would trigger an evaluation of the pending side effect?
I didn't check for this carefully though, just wanted to make sure that it works.

src/cuda_core.cpp
```diff
@@ -355,24 +356,38 @@ bool jitc_cuda_init() {
...
+            uint32_t driver_version_i = driver_to_ptx_isa_mappling[i][0],
+                     ptx_version_i    = driver_to_ptx_isa_mappling[i][1];
+
+            if (driver_version >= driver_version_i)
+                ptx_version = ptx_version_i;
+            else
+                break;
         }
 
+        device.ptx_version = ptx_version;
```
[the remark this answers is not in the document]
wjakob: It is set in line 354, no?

PR #144, opened 2025-05-05
Fix build with DRJIT_ENABLE_OPTIX=OFF

jeongseok-meta:
The build fails when `DRJIT_ENABLE_OPTIX=OFF` because the `optix.h` header is not included in that configuration. The OptiX function is also only used when this option is enabled:

https://github.com/mitsuba-renderer/drjit-core/blob/e63e186eefc2647e00efc2e1aeae3c66d996714e/src/io.cpp#L462-L464


dvicini:
is this still relevant? For me, building with DRJIT_ENABLE_OPTIX=OFF seems to work (on Linux)

jeongseok-meta:
Yes, it's still relevant so far. We're internally using another build system ([Buck2](https://buck2.build/)) instead of CMake, so there could be errors in configuring the build with it. Let me double-check this.

dvicini:
I am in the same situation as you but with Bazel/Blaze, but in the latest version didn't encounter such issues anymore.

dvicini:
Latest version that is used by drjit (1)

PR #145, opened 2025-05-19
Shader Execution Reordering

njroussel:
This PR adds support for Shader Execution Reordering (SER) in OptiX ray tracing calls (`jit_optix_ray_trace()`). It also adds an explicit reodering function if the user wants to reorder threads based on a key that does not involve any ray intersection information: `jit_optix_reorder`.

The SER functionality can be completely turned off with a new `ShaderExecutionReordering` JIT flag.

No new tests were added. The triangle test was updated to correctly adopt the new ray tracing interface.


wjakob:
LGTM!

wjakob:
(Thank you for your hard work, the implementation and description looks very polished.)

PR #146, opened 2025-05-27
Fix bug in symbolic CUDA loops

njroussel:
This PR is a fix for symbolic loops when state variables get assigned to other state variables, for example: 
```python
@dr.syntax
def f():
   previous, current = ...
    while cond(...): # both `previous` and `current` are state variables in this loop
        previous = current
        current = g(...)
        ...
```

In symbolic mode, with the CUDA backend, the variable getting assigned an other state variable (`previous` in the example above) would potentially be assigned the new value of the state variable (`g(...)` in the example)  rather than its original value (`current`'s value at the beginning of the loop).

When assembling a `LoopEnd` node in CUDA, for a loop with two state variable `a` and `b`, we'd generate something like this:
```
    mov.b32 %inner_in_a, %inner_out_a;
    mov.b32 %inner_in_b, %inner_out_b;
    bra l_cond;
l_done:
```

This is fine as long as the `inner_out` registers are not the same as any of the `inner_in` ones. However, for the original Python example, it would generate:
```
    mov.b32 %inner_in_a, %inner_out_a;
    mov.b32 %inner_in_b, %inner_in_a; // Wrong ! Should use `%inner_in_a` before it's updated
```

This PR adds temporary copies of state variables that get assigned to other variables, and uses these copies in the final assignment when needed. With this extra copy, we can guarantee that the final assignment will not use a value that has already been overridden.

Once this PR is merged, I'll add a test to `drjit`.

### DIFF
--- a/src/cuda_eval.cpp
+++ b/src/cuda_eval.cpp
@@ -55,6 +55,9 @@
 #if defined(DRJIT_ENABLE_OPTIX)
 #  include <drjit-core/optix.h>
 #endif
+#include <tsl/robin_set.h>
+
+using drjit::vector;
 
 // Forward declarations
 static void jitc_cuda_render(Variable *v);
@@ -377,6 +380,66 @@ static inline uint32_t jitc_fp16_min_compute_cuda(VarKind kind) {
             return 53;
     }
 }
+
+void jitc_cuda_render_loop_end(Variable *a0) {
+    const LoopData *ld = (LoopData *) a0->data;
+    uint32_t size = ld->size;
+
+    // Initialize scratch space
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+        in->scratch = out->scratch = 0;
+    }
+
+    // Mark inputs that will be read
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        in->scratch = 1;
+    }
+
+    /* If inner_out[i] == inner_in[j] (with i != j), we need to copy
+       `inner_in[j]` to a temporary to ensure that we can store its
+       original value to `inner_out[i]`, rather than its updated
+       value (i.e after the `inner_in[j] = inner_out[j]` assignment). */
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+
+        // If `out->scratch != 1` then we're certain that `out` is an input
+        if (in == out || !in->reg_index || !out->reg_index || out->scratch != 1)
+            continue;
+
+        fmt("    .reg.$b $v_tmp;\n"
+            "    mov.$b $v_tmp, $v;\n",
+            in, in, in, in, out);
+
+        // Mark that we have created a temporary
+        out->scratch = 2;
+    }
+
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        if (out->scratch == 2)
+            fmt("    mov.$b $v, $v_tmp;\n", in, in, in);
+        else
+            fmt("    mov.$b $v, $v;\n", in, in, out);
+    }
+
+    fmt("    bra l_$u_cond;\n\n"
+        "l_$u_done:\n",
+        a0->reg_index, a0->reg_index);
+}
+
 
 static void jitc_cuda_render(Variable *v) {
     const char *stmt = nullptr;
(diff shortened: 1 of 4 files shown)


wjakob:
Hi Nicolas,

this is pretty horrible, thank you for tracking down this issue and fixing it. In terms of the added data structures, the solution is a bit heavy, and I wonder if the same cannot also be accomplished by some use of internal fields of the `Variable` data structure. Please see here for a patch relative to yours, it also bundles in a fix for an unrelated issue I stumbled into.

```diff
diff --git a/src/cuda_eval.cpp b/src/cuda_eval.cpp
index 1359e8d..54459a2 100644
--- a/src/cuda_eval.cpp
+++ b/src/cuda_eval.cpp
@@ -381,6 +381,78 @@ static inline uint32_t jitc_fp16_min_compute_cuda(VarKind kind) {
     }
 }
 
+void jitc_cuda_render_loop_end(Variable *a0) {
+    const LoopData *ld = (LoopData *) a0->data;
+    uint32_t size = ld->size;
+
+    // Initialize scratch space
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+        in->scratch = out->scratch = 0;
+    }
+
+    // Mark inputs that will be read
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        in->scratch = 1;
+    }
+
+    /* If inner_out[i] == inner_in[j] (with i != j), we need to copy
+       `inner_in[j]` to a temporary to ensure that we can store its
+       original value to `inner_out[i]`, rather than its updated
+       value (i.e after the `inner_in[j] = inner_out[j]` assignment). */
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        fmt("    .reg.$b $v_tmp;\n"
+            "    mov.$b $v_tmp, $v;\n",
+            in, in, in, in, out);
+        // Mark that we have created a temporary
+        in->scratch = 2;
+    }
+
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        fmt("    .reg.$b $v_tmp;\n"
+            "    mov.$b $v_tmp, $v;\n",
+            in, in, in, in, out);
+
+        // Mark that we have created a temporary
+        in->scratch = 2;
+    }
+
+    for (uint32_t i = 0; i < size; ++i) {
+        Variable *in  = jitc_var(ld->inner_in[i]),
+                 *out = jitc_var(ld->inner_out[i]);
+
+        if (in == out || !in->reg_index || !out->reg_index)
+            continue;
+
+        if (in->scratch == 2)
+            fmt("    mov.$b $v, $v_tmp;\n", in, in, in);
+        else
+            fmt("    mov.$b $v, $v;\n", in, in, out);
+    }
+
+    fmt("    bra l_$u_cond;\n\n"
+        "l_$u_done:\n",
+        a0->reg_index, a0->reg_index);
+}
+
+
 static void jitc_cuda_render(Variable *v) {
     const char *stmt = nullptr;
     Variable *a0 = v->dep[0] ? jitc_var(v->dep[0]) : nullptr,
@@ -898,59 +970,11 @@ static void jitc_cuda_render(Variable *v) {
                 "l_$u_body:\n", a1, a0->reg_index, a0->reg_index);
             break;
 
-        case VarKind::LoopEnd: {
-                const LoopData *ld = (LoopData *) a0->data;
-
-                vector<Variable*> inner_in;
-                vector<Variable*> inner_out;
-                tsl::robin_set<Variable*> inner_in_set;
-
-                inner_in.reserve(ld->size);
-                inner_out.reserve(ld->size);
-                inner_in_set.reserve(ld->size);
-
-                for (uint32_t i = 0; i < ld->size; ++i) {
-                    uint32_t in_idx = ld->inner_in[i];
-                    Variable* in = jitc_var(in_idx);
-                    uint32_t out_idx = ld->inner_out[i];
-                    Variable* out = jitc_var(out_idx);
-
-                    // Which state variables are actually necessary ?
-                    if (in != out && in->reg_index && out->reg_index) {
-                        inner_in_set.insert(in);
-                        inner_in.emplace_back(in);
-                        inner_out.emplace_back(out);
-                    }
-                }
-
-                /* If inner_out[i] == inner_in[j] (with i != j), we need to copy
-                   `inner_in[j]` to a temporary to ensure that we can store its
-                   original value to `inner_out[i]`, rather than its updated
-                   value (i.e after the `inner_in[j] = inner_out[j]` assignment). */
-                vector<uint32_t> requires_copy;
-                requires_copy.reserve(ld->size);
-                for (uint32_t i = 0; i < inner_out.size(); ++i) {
-                    if (inner_in_set.find(inner_out[i]) != inner_in_set.end())
-                        fmt("    .reg.$b $v_tmp;\n"
-                            "    mov.$b $v_tmp, $v;\n",
-                            inner_in[i], inner_in[i],
-                            inner_in[i], inner_in[i], inner_out[i]);
-                }
-                for (uint32_t i = 0; i < inner_out.size(); ++i) {
-                    if (inner_in_set.find(inner_out[i]) != inner_in_set.end())
-                        fmt("    mov.$b $v, $v_tmp;\n",
-                            inner_in[i], inner_in[i], inner_in[i]);
-                    else
-                        fmt("    mov.$b $v, $v;\n",
-                            inner_in[i], inner_in[i], inner_out[i]);
-                }
-
-                fmt("    bra l_$u_cond;\n\n"
-                    "l_$u_done:\n",
-                    a0->reg_index, a0->reg_index);
-            }
+        case VarKind::LoopEnd:
+            jitc_cuda_render_loop_end(a0);
             break;
 
+
         case VarKind::LoopPhi:
             if (v->is_array())
                 v->reg_index = a3->reg_index;
diff --git a/src/internal.h b/src/internal.h
index 4467231..b12faba 100644
--- a/src/internal.h
+++ b/src/internal.h
@@ -224,8 +224,8 @@ struct alignas(64) Variable {
     /// How many times has this variable entry been (re-) used?
     uint32_t counter;
 
-    /// Unused
-    uint32_t unused;
+    /// Free-use field temporarily used by various parts of Dr.Jit
+    uint32_t scratch;
 
     // ================  Essential flags used in the LVN key  =================
     // (+15 bits)
diff --git a/src/lock.h b/src/lock.h
index ed48245..3bd0182 100644
--- a/src/lock.h
+++ b/src/lock.h
@@ -33,6 +33,8 @@ inline void lock_acquire(Lock &lock) { lock.lock(); }
 inline void lock_release(Lock &lock) { lock.unlock(); }
 #endif
 
+extern void jitc_sanitation_checkpoint();
+
 /// RAII helper for scoped lock acquisition
 class lock_guard {
 public:
```

njroussel:
:+1: 

Your patch was a bit too aggressive (or maybe it was just a typo/bug), it would always create a temporary. I pushed a new version with some minor modifications such that it only does it when necessary.

wjakob:
Awesome, thanks for the fix!

PR #147, opened 2025-06-02
Use `JitFlag::Debug` to turn on OptiX validation mode

njroussel:
This PR removes the OptiX validation mode as a CMake option, and instead adds it as a feature of `JitFlag::Debug`.

PR just for the CI.


wjakob:
LGTM 👍

wjakob (approved):

PR #151, opened 2025-06-11
Improve packet scatters and gathers

DoeringChristian:
This PR enables us to render a specialized version of the PacketScatter op in CUDA when a reduction operation is specified, with f16 types. It uses the `red.global.add.noftz.f16x2` instruction.

### DIFF
--- a/src/cuda_packet.cpp
+++ b/src/cuda_packet.cpp
@@ -10,10 +10,13 @@
 
 #include "eval.h"
 #include "cuda_eval.h"
-#include "llvm_packet.h"
 #include "var.h"
 #include "op.h"
 #include "log.h"
+
+static const char *reduce_op_name[(int) ReduceOp::Count] = {
+    "", "add", "mul", "min", "max", "and", "or"
+};
 
 void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
                                     const Variable *index, const Variable *mask) {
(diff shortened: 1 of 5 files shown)


wjakob:
Potentially relevant: `sm_90`+ support `red.global.add.v2.f16x2` and `red.global.add.v4.f16x2` (so 8 FP16 accumulations in one instruction). This should be available on RTX5090.

src/cuda_packet.cpp
```diff
@@ -121,13 +125,67 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+    const Variable *v0 = jitc_var(values[0]);
+
+    uint32_t count = (uint32_t) values.size(),
+             tsize = type_size[v0->type];
+
+    put("    {\n");
+
+    fmt("        .reg.f16x2 %tmp;\n");
+
+    fmt("        mad.wide.$t %rd3, $v, $u, $v;\n",
```
merlinND, line 147:
```suggestion
    put("    {\n"
        "        .reg.f16x2 %tmp;\n"
        "        mad.wide.$t %rd3, $v, $u, $v;\n",
```

src/cuda_packet.cpp
```diff
@@ -121,13 +125,67 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+        index, index, tsize, ptr);
+
+    if (v0->type != (uint32_t)VarType::Float16)
+        jitc_fail("Packeted scatter reductions are only supported with f16 "
+                  "variables.");
```
merlinND, line 152: Move this check earlier + include the name of the type you actually got in the error message (`type_name[v0->type]` I think)

src/cuda_packet.cpp
```diff
@@ -15,6 +15,10 @@
 #include "op.h"
 #include "log.h"
 
+static const char *reduce_op_name[(int) ReduceOp::Count] = {
```
merlinND, line 18: Would it be possible to not re-define this? (vs `cuda_scatter.cpp`).
Having it duplicated in separate locations makes it error-prone when the enum changes.
Although I see it's already defined another time in `llvm_scatter.cpp`, so maybe it's not easy to have it only once.

src/cuda_packet.cpp
```diff
@@ -121,13 +125,67 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+    fmt("        mad.wide.$t %rd3, $v, $u, $v;\n",
+        index, index, tsize, ptr);
+
+    if (v0->type != (uint32_t)VarType::Float16)
+        jitc_fail("Packeted scatter reductions are only supported with f16 "
+                  "variables.");
+
+    const char *op = reduce_op_name[(uint32_t) psd->op];
+
+    if (count % 2 == 0) {
```
merlinND, line 156: I think it would be great to expand this to use the `.v2, .v4, .v8` variants of `red`, like @wjakob suggested.

src/cuda_packet.cpp
```diff
@@ -121,13 +125,67 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+/**
+ * Render the code required to scatter reduce a packet of variables.
+ */
+void jitc_cuda_render_scatter_reduce_packet(const Variable *v,
+                                            const Variable *ptr,
+                                            const Variable *index,
+                                            const Variable *mask) {
+    bool is_masked = !mask->is_literal() || mask->literal != 1;
+    PacketScatterData *psd = (PacketScatterData *) v->data;
+    const std::vector<uint32_t> &values = psd->values;
```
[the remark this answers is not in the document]
wjakob: It's fine to use std::vector. I just prefer `dr::vector` in header files, so that we don' t have to pull in STL code everywhere.

src/cuda_packet.cpp
```diff
@@ -121,13 +124,117 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+
+    uint32_t count = (uint32_t) values.size(),
+             tsize = type_size[v0->type];
+
+    if (count % 2 != 0)
+        jitc_fail("jitc_cuda_render_scatter_reduce_packet(): Number of "
+                  "elements not supported for reduction.");
+
+    if (ts->compute_capability >= 90) {
+        // Use the new `red.global.v2` instructions. This enables both min & max
```
wjakob, line 149: Is this comment up-to-date? `.v2` is only two elements, and we're doing wider ones AFAIK.

src/cuda_packet.cpp
```diff
@@ -121,13 +124,117 @@ void jitc_cuda_render_gather_packet(const Variable *v, const Variable *ptr,
...
+        // as well as packet reductions with larger packet sizes per iteration
+        // and `f32` types.
+        const uint32_t MAX_VARS_PER_IT = 16 / tsize;
+
+        // Find largest power of two dividing the number of variables, smaller
```
wjakob, line 154: Text is oddly formatted

src/op.cpp
```diff
@@ -1905,11 +1905,11 @@ void jitc_var_gather_packet(size_t n, uint32_t src_, uint32_t index, uint32_t ma
     auto [var_info, index_v, mask_v] =
         jitc_var_check("jit_var_gather_packet", index, mask);
 
-    if ((n & (n-1)) || n == 1)
+    if (n == 1)
         jitc_raise("jitc_var_gather_packet(): vector size must be a power of two "
```
wjakob, line 1909: Error message seems out of date now.

src/op.cpp
```diff
@@ -1905,11 +1905,11 @@ void jitc_var_gather_packet(size_t n, uint32_t src_, uint32_t index, uint32_t ma
...
                    "and >= 1 (got %zu)!", n);
 
-    if ((src_info.size & (n-1)) != 0 && src_info.size != 1)
+    if (src_info.size % 2 != 0 && src_info.size != 1)
         jitc_raise("jitc_var_gather_packet(): source r%u has size %u, which is not "
```
wjakob, line 1913: Error message seems out of date now.

src/op.cpp
```diff
@@ -2639,9 +2650,29 @@ uint32_t jitc_var_scatter_packet(size_t n, uint32_t target_,
...
+                    use_packet_op &&
+                    ((targe_type == VarType::Float32 && op == ReduceOp::Add) ||
+                     (targe_type == VarType::Float16 &&
+                      (op == ReduceOp::Add || op == ReduceOp::Min ||
+                       op == ReduceOp::Max)));
         }
     }
 
+    // If the packet size is not divisible by two we cannot use packet ops.
+    use_packet_op = use_packet_op && n > 1 && n % 2 == 0;
```
wjakob, line 2674: If we move the logic to code generation, perhaps it's easier to not even have special handling for the n=1 case here.

wjakob (commented):
This is cool, some feedback from me.  The first two are from an earlier partial review and may no longer apply. (It says the files are "Outdated").

PR #152, opened 2025-06-12
Fixes for 50-series GPUs (CUDA backend)

merlinND:
## Reductions

For f16 or f16x2, only `.add` is supported by the `red` instruction.
Mark other reduction operations as unsupported.

## Compress / prefix reduction

Symptom: on 50-series GPUs, any usage of `dr::compress()` with an input of size larger than 4096 would hang with 100% GPU usage.

After looking at [this Cub implementation](https://github.com/NVIDIA/cccl/blob/073703ff06f4162f0d59a2dd14689de073f0df24/cub/cub/agent/agent_radix_sort_onesweep.cuh#L282-L286), it seems that our implementations of the decoupled-lookback pattern in the Compress and Prefix Reduction kernels needed a `__threadfence_block()`.


resources/compress.cuh
```diff
@@ -107,6 +107,9 @@ KERNEL void compress_large(const uint8_t *in, uint32_t *out, uint64_t *scratch,
        Based on "Single-pass Parallel Prefix Scan with Decoupled Look-back"
        by Duane Merrill and Michael Garland */
     while (true) {
+        // Avoid re-ordering of loads and stores under the relaxed memory model.
+        __threadfence_block();
```
merlinND, line 111: I have not tested the performance impact of this change, and I'm not 100% certain that:
- This is exactly the kind of fence we need.
- It is placed in the right location.

Same for the other fence in `block_prefix_reduce()`.

src/op.cpp
```diff
@@ -2231,9 +2231,8 @@ bool jitc_can_scatter_reduce(JitBackend backend, VarType vt, ReduceOp op) {
             if (op == ReduceOp::Add && compute_capability < 60)
                 return false;
 
-            // Half-precision min/max reductions require sm_90
-            if ((op == ReduceOp::Min || op == ReduceOp::Max) &&
-                compute_capability < 90)
+            // Half-precision reductions other than `add` are not supported on CUDA
+            if (is_cuda && op != ReduceOp::Add)
```
merlinND, line 2235: I have left the implementation in `cuda_scatter.cpp` intact.
Ideally, we would be able to still support those reductions with a workaround.

merlinND: Actually I think changing this function is not enough to clearly indicate to the user that this reduction is not supported.

resources/compress.cuh
merlinND: → replaced with `volatile` loads

PR #153, opened 2025-06-19
Frozen Functions Kernel History

DoeringChristian:
Enables frozen functions to generate kernel history entries, when replaying kernels.


merlinND (commented):
This looks nice, good job catching that this was missing!

PR #154, opened 2025-06-21
Frozen functions: Avoid invalid memory read for literals

dvicini:
This fixes a small issue in record_ts: For literal variables, the `slot` field is undefined. Therefore, we need to make sure to only use `info.slot` if we are not in the case where `type == ParamType::Register`.

The current code likely "works" due to zero initialization of memory, but doesn't work when running with a sanitizer that explicitly does not zero initialize

### DIFF
--- a/src/record_ts.cpp
+++ b/src/record_ts.cpp
@@ -2250,11 +2250,11 @@ bool RecordThreadState::has_variable(const void *ptr) {
  * accessed.
  */
 void RecordThreadState::add_param(AccessInfo info) {
-    RecordedVariable &rv = m_recording.recorded_variables[info.slot];
     if (info.type == ParamType::Output) {
 
         jitc_log(LogLevel::Debug, " <- param s%u", info.slot);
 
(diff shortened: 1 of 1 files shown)


dvicini:
Ah, it's possible that the bug is at the call site then?

This is the code I am looking at:
```cpp
AccessInfo info;
std::memcpy(&info.extra.data, &p.src, sizeof(uint64_t));
info.extra.offset    = p.offset;
info.extra.type_size = p.size;
info.type            = ParamType::Register;
info.pointer_access  = false;
add_param(info);
```

This is around line ~1880 in record_ts.cpp

DoeringChristian:
Ah right. I didn't think of that special case. This is the only case, where the RecordThreadState, actually has to handle literals (i.e. when they are mapped to a vcall). In this case, your suggestion makes the most sense, moving the lookups into the conditionals.

dvicini:
Okay, should I go ahead and merge then? You are the authority on this code, so I will trust your judgment

DoeringChristian:
Ok, then please go ahead. Thanks for checking!

PR #156, opened 2025-07-18
Add ``jit_var_mul_wide()`` operation for wide integer multiplication

wjakob:
This PR adds a wide (e.g. i32xi32 -> i64) integer multiplication node, which is the central operation underlying the Philox pseudorandom number generator.

Test coverage is added separately at the Dr.Jit level.


njroussel (approved):
LGTM

PR #159, opened 2025-07-30
Fix performance cliffs in `jitc_var_reindex()`

wjakob:
The operation `dr.gather()` can inline an expression tree by rewriting the underlying indices. However, this exposes some dangerous performance cliffs:

1. If an variable is referenced many times within an expression, Dr.Jit will repeatedly rewrite it (and its dependencies), which can lead to explosive cost.

2. Repeated + nested gathering (e.g., when building a MIP map) can lead to exponential growth of the expressions.

This commit fixes both issues by

1. Introducing a temporary map that memoizes rewritten nodes so that we only generate them once.

2. Rejecting expressions that are too big. I've chosen an arbitrary threshold of 200 variables for this.

I verified that each of these fixes alone suffices to address a recent issue where dr.gather calls became too expensive.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -1608,8 +1608,16 @@ static uint32_t jitc_scatter_gather_index(uint32_t source, uint32_t index) {
 
 /// Change all indices/counters in an expression tree to 'new_index'
 static uint32_t jitc_var_reindex(uint32_t var_index, uint32_t new_index,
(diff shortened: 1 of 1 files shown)


njroussel (approved):
LGTM
That's the fix I had in mind too.

I don't really have an opinion on the threshold of 200 variables, it seems sensible.

PR #160, opened 2025-08-06
Support for excess register spilling to CUDA shared memory

lnuic:
This PR introduces a new JIT flag JitFlag::SpillToSharedMemory to enable spilling of excess registers into shared memory when register pressure is high.

Spilling to shared memory can reduce latency compared to spilling into local memory, potentially improving performance in register-heavy CUDA kernels. The feature is relevant only on backends that support this behavior (e.g., CUDA with PTX ISA 8.7+ and sm_75 or higher).

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1617,6 +1617,9 @@ enum class JitFlag : uint32_t {
     /// Reorder threads in OptiX after a ray-intersection
     ShaderExecutionReordering = 1 << 23,
 
+    /// Spill registers under pressure into shared memory
+    SpillToSharedMemory = 1 << 24,
+
     /// Default flags
     Default = (uint32_t) ConstantPropagation | (uint32_t) ValueNumbering |
               (uint32_t) FastMath | (uint32_t) SymbolicLoops |
@@ -1624,7 +1627,8 @@ enum class JitFlag : uint32_t {
               (uint32_t) MergeFunctions | (uint32_t) OptimizeCalls |
               (uint32_t) SymbolicConditionals | (uint32_t) ReuseIndices |
               (uint32_t) ScatterReduceLocal | (uint32_t) PacketOps |
-              (uint32_t) KernelFreezing | (uint32_t) ShaderExecutionReordering,
+              (uint32_t) KernelFreezing | (uint32_t) ShaderExecutionReordering |
+              (uint32_t) SpillToSharedMemory,
 
     // Deprecated aliases, will be removed in a future version of Dr.Jit
     LoopRecord = SymbolicLoops,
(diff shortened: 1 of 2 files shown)


wjakob:
Interesting! Did you see some speedups from this?

How about naming this ``SpillToSharedMemory``?

wjakob:
In the PTX reference, this is specified as a function-level attribute:
```
.entry foo (...)
.pragma "enable_smem_spilling";   // Enable shared memory spilling for this function
{
    ...
}```

**Comment by lnuic** [comment]
> Interesting! Did you see some speedups from this?

From the small tests I've run, I didn't notice any significant performance improvements. Neither with Dr.Jit using custom register-heavy synthetic test nor with Mitsuba on the gallery scenes.

> In the PTX reference, this is specified as a function-level attribute:
> 
> ```
> .entry foo (...)
> .pragma "enable_smem_spilling";   // Enable shared memory spilling for this function
> {
>     ...
> }
> ```
> 
> Not sure if this is needed.

Using enable_smem_spilling as a function-level attribute causes a compilation error, so this may be a mistake in the documentation:

```.version 9.0
.target sm_89
.address_size 64

.entry drjit_fff0b79c8b2075f9574e533c8acc1907(.param .align 8 .b8 params[32])
.pragma "enable_smem_spilling";
{
    .reg.b8  %b <7>; .reg.b16  %w<7>; .reg.b32 %r<7>;
    .reg.b64 %rd<7>; .reg.f16  %h<7>; .reg.f32 %f<7>;
    .reg.f64 %d <7>; .reg.pred %p<7>;

    mov.u32 %r0, %ctaid.x;
    mov.u32 %r1, %ntid.x;
    mov.u32 %r2, %tid.x;
    mad.lo.u32 %r0, %r0, %r1, %r2;
    ld.param.u32 %r2, [params];
    setp.ge.u32 %p0, %r0, %r2;
    @%p0 ret;

body: // sm_89
    ld.param.u64 %rd0, [params+8];
    mad.wide.u32 %rd0, %r0, 4, %rd0;
    ld.global.cs.b32 %f4, [%rd0];
    ld.param.u64 %rd0, [params+16];
    mad.wide.u32 %rd0, %r0, 4, %rd0;
    ld.global.cs.b32 %f5, [%rd0];
    add.ftz.f32 %f6, %f4, %f5;
    ld.param.u64 %rd0, [params+24];
    mad.wide.u32 %rd0, %r0, 4, %rd0;
    st.global.cs.b32 [%rd0], %f6;
    ret;
}


ptxas application ptx input, line 6; error   : Pragma 'enable_smem_spilling' is allowed only within function scope.
```

Furthermore, it seems that this pragma only applies to entry functions, as using it inside a device function causes the following error:
```
ptx2llvm-module-001, line 286; error   : Pragma 'enable_smem_spilling' cannot be specified for device functions
```

njroussel (approved):
Thanks, LGTM

PR #161, opened 2025-08-11
Fixed literal target for jitc_coop_vec_outer_product_accum

DoeringChristian:
This PR fixes the case where the target buffer of `jitc_coop_vec_outer_product_accum` is a literal instead of a buffer or nothing, by returning the opaque version of the target, in which the operation accumulates. The fix is inspired by the `jitc_var_scatter` function. This fixes a bug where gradients of neural networks would not be calculated correctly, if the previous gradient was a non zero sized literal, for example wen setting it in beforehand.


wjakob:
Awesome, thank you!

PR #163, opened 2025-08-18
Better Int8/UInt8 support

merlinND:


src/cuda_eval.cpp
```diff
@@ -339,6 +339,7 @@ void jitc_cuda_assemble_func(const CallData *call, uint32_t inst,
                     "    setp.ne.u16 $v, %w0, 0;\n",
                     offset, v);
         } else if (v->is_literal()) {
+            // We need a special case for subword-sized types
```
merlinND, line 342: I think this comment is no longer relevant?

PR #168, opened 2025-08-21
Add LLVM API initialization failed warning

Speierers:
This patch adds a warning message when LLVM initialization fails.


wjakob:
Thanks!

PR #172, opened 2025-08-29
Fix minor typing issues in gl_interop

dvicini:
This fixes compilation errors for the case when `DRJIT_DYNAMIC_CUDA` is off. In that case, we have to be more careful to use the cuda types such as `CUresult` or `CUdeviceptr` where needed.


wjakob:
Thanks!

PR #174, opened 2025-09-08
Event API

wjakob:
This PR adds a new event API to Dr.Jit-core.

This functionality directly maps to standard CUDA events on the CUDA backend. The LLVM backend provides the same interface via the nanothread thread pool.

This feature makes it possible to:

1. Record an event (recording multiple times invalidates the old event)

2. Query the event for completion without blocking

3. Blocking to wait for event completion. This is especially useful to wait for an event scheduled some time ago, which can be much more efficient than ``jit_sync_thread/device()`` when other work was scheduled in the meantime.

4. Measuring the relative time between two events. This provides a cheaper alternative to the kernel history for simple benchmarking tasks.

5. Extracting the native event pointer for advanced use cases

Discussion:

- The event API does not do anything clever with regards to kernel freezing at the moment. It was not clear to me if this even makes sense. Events cannot be fused into kernels, and they don't produce outputs or consume array inputs. How would one replay events, and where would they be stored? So I am assuming this is a feature that will be used outside of frozen functions.

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -2917,3 +2917,97 @@ extern JIT_EXPORT uint32_t jit_coop_vec_outer_product_accum(
     const MatrixDescr *descr,
     uint32_t a,
     uint32_t b);
+
+// ====================================================================
+//                            Event API
+// ====================================================================
+
+/// Opaque handle type for events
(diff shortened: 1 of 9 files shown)


njroussel (approved):
LGTM

PR #177, opened 2025-09-24
New JIT nodes for `dr.scatter_cas()` and `dr.scatter_exch()`

njroussel:
This PR lays the groundwork for implementing `dr.scatter_cas()` and `dr.scatter_exch()`.

The implementation shares a lot of similarities with `scatter_inc`, as the outputs are only valid when used in the same kernel as where the memory write takes place. To do so, it also relies on the `consumed` field of `Variables`.

PTX has instructions to peform both of these operations trivially, where as the LLVM implementation must rely on a loop over the vectorization width; there currently is no `cmpxchg` or ` atomicrmw` for vectorized types in the LLVM IR spec.

As always, tests are contained in the Dr.Jit PR: https://github.com/mitsuba-renderer/drjit/pull/450

### DIFF
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -1053,6 +1053,31 @@ extern JIT_EXPORT uint32_t jit_var_scatter_inc(uint32_t *target,
                                                uint32_t mask);
 
 /**
+ * \brief Perform an atomic compare-and-swap
+ *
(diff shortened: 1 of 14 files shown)


wjakob (commented):
Just a small request, otherwise this looks great. It makes sense that CAS is more complicated than the other scatter operations.

PR #189, opened 2026-04-29
[WIP] Add Metal backend to Dr.Jit Core

Speierers:
# Drjit Core: Metal (Apple Silicon GPU) backend

This PR adds a **Metal compute backend** to Drjit Core, enabling JIT
execution on Apple Silicon GPUs (M-series, Metal 3 / macOS 13+). The
backend is a sibling of the existing CUDA backend and is selected via the
new `JitBackend::Metal`.

It is the foundation for the Mitsuba and Drjit Metal variants
(see the companion PRs in the `mitsuba3` and `drjit` repositories).

## Summary

- Apple platforms get a first-class GPU JIT backend that does not rely
  on CUDA / OptiX.
- Build / link / runtime cost on non-Apple platforms is **zero**: every
  Metal source file is gated on `DRJIT_ENABLE_METAL` and every Metal
  framework dependency is wrapped in an `if (APPLE AND DRJIT_ENABLE_METAL)`
  block in CMake.
- The CUDA and LLVM backends are unchanged.

## How the backend works (short version)

* **Codegen.** Recorded Drjit graphs are lowered to a single
  **Metal Shading Language** compute kernel as text, by
  `src/metal_eval.cpp`. The format-string conventions mirror
  `src/cuda_eval.cpp` (`$v`, `$t`, ...), adapted to MSL syntax.
* **Compiler.** MSL source is compiled at run time via Apple's
  `MTLDevice::newLibrary(NS::String*, MTLCompileOptions*, &err)`. Compiled
  pipelines (`MTL::ComputePipelineState`) are cached on a hash of the
  source text, so repeated launches of the same recorded graph reuse the
  existing pipeline.
* **Launch.** Inputs and outputs are passed by **GPU virtual address**
  (`MTL::Buffer::gpuAddress()`) in a parameter buffer bound at
  `[[buffer(0)]]`. Each accessed buffer is marked with `useResource()` on
  the encoder so Metal's residency system makes the pages resident, then
  the kernel is dispatched via `dispatchThreads` (or `dispatchThreadgroups`
  for utility kernels).
* **Utility kernels** (block reductions, prefix sums, compress, mkperm,
  expand-reduce atomics, aggregate copy) live as MSL source in
  `resources/metal_kernels.metal`, compiled once at backend init into a
  shared `MTL::Library`.
* **Ray tracing.** The Metal backend exposes `jit_metal_configure_scene`
  and `jit_metal_ray_trace` (in `include/drjit-core/metal.h`). The trace
  IR node carries a per-scene JIT-variable handle, so kernels can bind the
  correct TLAS / intersection-function table per trace operation.

### New files (~5,500 LoC)

| File | Counterpart | Purpose |
|------|-------------|---------|
| `src/metal.h`, `src/metal_core.cpp` | `cuda.h`, `cuda_core.cpp` | Backend init, device enumeration, MSL → pipeline compilation, scene lifetime |
| `src/metal_api.{h,cpp}` | `cuda_api.{h,cpp}` | Public C API (`jit_metal_*`) |
| `src/metal_eval.{h,cpp}` | `cuda_eval.{h,cpp}` | MSL kernel codegen (largest file, ~1.4 k LoC) |
| `src/metal_array.{h,cpp}` | `cuda_array.{h,cpp}` | Per-variable creation / dispatch |
| `src/metal_scatter.{h,cpp}` | `cuda_scatter.{h,cpp}` | Atomic scatter / reductions |
| `src/metal_ts.{h,cpp}` | `cuda_ts.{h,cpp}` | `MetalThreadState`, kernel launch, command buffers (~1.4 k LoC) |
| `src/metal_stage_buffer.{h,cpp}` | (CUDA pinned mem) | Stage-buffer abstraction (Private vs Shared storage) |
| `src/metal_coop_vec.{h,cpp}` | `cuda_coop_vec.*` | Cooperative-vector lowering |
| `src/metal_mps.mm` | (cuBLAS) | MPS-backed `batched_gemm` |
| `include/drjit-core/metal.h` | `optix.h` | Public API |
| `resources/metal_kernels.metal` | (CUDA utility .cu) | Precompiled utility kernels |

### Touched shared code

Small "add a Metal branch next to the CUDA branch" edits in:
`init.cpp`, `var.cpp`, `eval.cpp`, `op.cpp`, `call.cpp`, `record_ts.cpp`,
`malloc.cpp`, `io.cpp`, `internal.h`, `api.cpp`.

### Public API surface

```c
// include/drjit-core/metal.h
void *  jit_metal_context();           // MTL::Device*
void *  jit_metal_command_queue();     // MTL::CommandQueue*

uint32_t jit_metal_configure_scene(
    void *tlas, void **resources, uint32_t n_resources,
    void *intersection_fn_library,
    uint32_t n_ift_entries,
    const char **ift_function_names,
    void **ift_buffers, const uint32_t *ift_buffer_slots,
    uint32_t geometry_types_mask);

void    jit_metal_ray_trace(uint32_t n_args, uint32_t *args,
                            uint32_t mask, uint32_t *out, uint32_t n_out,
                            uint32_t scene);

void *  jit_metal_lookup_buffer(void *ptr, size_t *offset);
void    jit_metal_invalidate_scene_tlas(uint32_t scene_index);
```

`jit_metal_lookup_buffer` lets host code resolve a JIT-managed pointer
back to its `MTL::Buffer*` (Metal binds resources by buffer + offset, not
raw GPU address). `jit_metal_invalidate_scene_tlas` is a frozen-function
safety hook.

## New dependencies (Apple only)

All gated on `DRJIT_ENABLE_METAL` so non-Apple builds are unaffected.

* **`metal-cpp`** — Apple's official C++ wrapper around the Objective-C
  Metal API. Vendored as a placeholder under `ext/metal-cpp/`; the
  user downloads `metal-cpp_macOS15_iOS18.zip` (or newer) from
  <https://developer.apple.com/metal/cpp/> and extracts the
  `Foundation/`, `Metal/`, and `QuartzCore/` directories. Header-only.
* **System frameworks** linked into `drjit-core`:
  * `Foundation`, `Metal`, `MetalPerformanceShaders`, `QuartzCore`.
* **Objective-C++ ARC.** The `metal-cpp` headers and `metal_mps.mm`
  require Objective-C++ compilation with ARC. The CMake recipe sets the
  appropriate file-properties for the Metal sources.

No new Python or third-party-source dependencies.

## Build

```cmake
# Default off; opt in on Apple platforms with:
cmake -DDRJIT_ENABLE_METAL=ON ..
```

The CMake block lives at the bottom of `CMakeLists.txt` next to the
existing OptiX block:

```cmake
if (APPLE AND DRJIT_ENABLE_METAL)
  target_compile_definitions(drjit-core PRIVATE -DDRJIT_ENABLE_METAL=1)
  target_compile_definitions(drjit-core PRIVATE
    -DDRJIT_METAL_KERNEL_SOURCE="${CMAKE_CURRENT_SOURCE_DIR}/resources/metal_kernels.metal")
  target_include_directories(drjit-core PRIVATE ext/metal-cpp)
  target_link_libraries(drjit-core PRIVATE
    "-framework Foundation"
    "-framework Metal"
    "-framework MetalPerformanceShaders"
    "-framework QuartzCore")
endif()
```

## Limitations and design decisions

| Topic | Decision | Reason |
|-------|----------|--------|
| Float64 | Auto-downcast to Float32 + one-time warning | Apple GPUs have no FP64 |
| FP32 atomic add | Routed through `ReduceMode::Expand` | Avoids precision loss from non-deterministic atomic ordering, matches CUDA accuracy |
| Fast math | Disabled for user kernels (enabled for utility kernels) | MSL fast-math eliminates `isnan` / `isinf` |
| Cooperative vectors | Per-thread scalar with `simdgroup_matrix` fast path | Metal has no equivalent of WMMA / CUTE |
| GEMM | Routed through Metal Performance Shaders | Only sane way to get tuned matmul on Metal |
| `ArrayMeta::backend` | Widened from 2 → 3 bits in the python-binding metadata | `JitBackend::Metal == 4` doesn't fit in 2 bits |

## Test plan

- [ ] All existing Drjit Core unit tests pass on macOS with
      `DRJIT_ENABLE_METAL=OFF` (no regression for Apple users who don't
      opt in).
- [ ] All existing Drjit Core unit tests pass on Linux / Windows
      (no regression for non-Apple platforms).
- [ ] On macOS with `DRJIT_ENABLE_METAL=ON`, the Metal-specific test
      coverage in the companion `drjit` PR runs green (kernel history,
      vcalls, frozen functions, cooperative vectors, ray tracing).


wjakob:
Thanks you very much for your PR Sébastien! I've done a thorough pass over it and changed many things. At this point, I think it is good enough to be merged (small fixes can be done out of band). Some of the refactoring needed to supporting 3 backends has also improved the other backends.

wjakob:
Worth noting: the macOS CI machine can't test this code since it is running an ancient version of macOS. That's on the TODO list and will be fixed.

PR #190, opened 2026-05-18
Butterfly warp-reduction for packet scatter-reduce

lnuic:
Adds a packet-aware butterfly path to `jitc_cuda_render_scatter_reduce_packet` that mirrors the existing per-element `_bfly_32` / `_bfly_64` intrinsics but processes all N channels of a packet through one match.any.sync, one butterfly tree and one set of leader atomics. The leader can issue either N scalar atomics or `red.global.vN` vector atomics (cc>=9.0, CUDA>=13.2 in OptiX).


wjakob:
Thanks!

PR #192, opened 2026-05-24
Several bug fixes 

lnuic:
See https://github.com/mitsuba-renderer/drjit/pull/503 for details.


wjakob:
These are all great catches, thank you for tracking them down Lovro!

PR #195, opened 2026-06-11
Metal: specify min macOS version for pre-built kernels

merlinND:
On macOS 15.5, DrJit would fail at import time due to the pre-built Metal library being compiled for a more recent deployment target:

```
>>> jit_metal_init(): could not instantiate the utility kernel library for device "Apple M3 Max": This library is using a deployment target (0x00020008) that is not supported on this OS.
```

This PR changes the resources `Makefile` to specify a minimum supported macOS version (13.0, chosen arbitrarily) and re-builds the library.
@wjakob please let me know if you would like to target a different minimum macOS version.


wjakob:
Thanks!

PR #197, opened 2026-06-12
Fix Metal readback race in scatter_inc/exch/cas

lnuic:
The write pointer of scatter_inc/exch/cas sat in the side-effect Nop's dep[3], which jitc_var_traverse() skips, so kernels declared the scatter target to Metal as read-only (useResource:), and a subsequent readback could overlap the still-running kernel and return a partial result. Move the pointer to the traversed dep[1] so the write is declared. CUDA/LLVM were unaffected, as their execution is currently implicitly synchronized by the in-order stream.

### DIFF
--- a/src/op.cpp
+++ b/src/op.cpp
@@ -2425,7 +2425,7 @@ uint32_t jitc_var_scatter_inc(uint32_t *target_p, uint32_t index, uint32_t mask)
                                        VarType::Void, var_info.size, symbolic,
                                        result, jitc_var(result)));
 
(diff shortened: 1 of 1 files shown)


wjakob:
I am not sure if that's the right fix. dep[3] is used to keep the source variable out of the program and kernel argument list (we're already passing a pointer to it). I have a separate fix for this (also ran into it) but have not pushed it.

wjakob:
Fixed via a55d6993385a002981fb03fd420be4f07f7158fd

PR #198, opened 2026-06-12
Fix loop codegen crash when two state variables share one update value

lnuic:
At the end of a symbolic loop, the CUDA and Metal backends use temporaries to break aliasing between state slots. The temporary was keyed on the destination register, so when two state variables take their next value from the same source, only the first temporary was declared and the second writeback referenced an undeclared register, ptxas then rejected the kernel (`Unknown symbol '%f10_tmp'`); similar error for MSL on Metal.

Fix: key the temporary on the source register, so all slots reading the same variable share one snapshot.

Minimal reproducer:

```python
​import drjit as dr
from drjit.cuda import Float, UInt32

@dr.syntax
def f():
    a, b, c = Float(1, 2), Float(3, 4), Float(5, 6)
    i = UInt32(0, 0)
    while i < 3:
        a, b, c = c, c + 0, a
        i += 1
    return a, b, c

f()  # before: ptxas fatal; after: ([5, 6], [5, 6], [1, 2])
```

### DIFF
--- a/src/cuda_eval.cpp
+++ b/src/cuda_eval.cpp
@@ -441,7 +441,7 @@ void jitc_cuda_render_loop_end(Variable *a0) {
 
         fmt("    .reg.$b $v_tmp;\n"
             "    mov.$b $v_tmp, $v;\n",
-            in, in, in, in, out);
+            out, out, out, out, out);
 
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

PR #199, opened 2026-06-13
Fix half-precision Min/Max reduction and Infinity constant

lnuic:
1. **LLVM Min/Max return 0 for Float16** 
 the reduction seeds its accumulator with `numeric_limits<half>::infinity()`, but `half` has no `numeric_limits` specialization, so the identity collapsed to `0` instead of ±inf. Fixed by using `float` value of the ±inf (matching what Add/Mul already do).
3. **`Infinity<half>` was −inf** (`0xfc00` → `0x7c00`)  
 broke `isinf`/`isfinite` for half in C++ scalar/packet mode.
5. **CUDA Float16 Min identity was max-finite** 
 (`0x7BFF` → `0x7C00`)  a block whose minimum is +inf returned 65504 instead of +inf.

## Reproducers

```python
import drjit as dr
inf = float("inf")

# Bug 1 (LLVM): min/max of Float16 return 0
from drjit.llvm import Float16 as LF16
print(dr.min(LF16(3, 1, 4))[0], dr.max(LF16(-3, -1))[0])   # correct: 1, -1   (before this fix: 0, 0)

# Bug 2 (scalar half): isinf/isfinite always wrong
from drjit.scalar import Array3f16
print(dr.isinf(Array3f16(1, inf, -inf)))                   # correct: [F,T,T]   (before this fix: [F,F,F])

# Bug 3 (CUDA): min of an all-+inf block clamps to 65504
from drjit.cuda import Float16 as CF16
print(dr.min(CF16(inf, inf, inf))[0])                      # correct: inf   (before this fix: 65504)


wjakob:
Not sure I understand point 1: why would we need a higher precision for minimum/maximum? It seems to me that fp16 has all that is needed to compute that reliably.

lnuic:
Because we use a custom half class, querying [std::numeric_limits<Value>::infinity()](https://github.com/mitsuba-renderer/drjit-core/blob/641be40f5ecdccb0aeb1acc0fdaa676b9b4295f3/src/llvm_red.h#L33) returns 0: there is no std::numeric_limits specialization for drjit::half, so the primary template is used and infinity() value-initializes to half(0). That is not the desired behavior here.

There are a few ways to fix this: we could add the necessary std::numeric_limits specialization for half so it works directly, hardcode the infinity constant, or use float as the accumulator. I chose the float approach as it requires the least code and we use the [same pattern for RedMul and RedAdd](https://github.com/mitsuba-renderer/drjit-core/blob/641be40f5ecdccb0aeb1acc0fdaa676b9b4295f3/src/llvm_red.h#L11).

Also, when we compute the min we use std::min, which compares with [<](https://github.com/mitsuba-renderer/drjit-core/blob/371f1453e1e3b87b450cace39816f151e8fe7b8c/include/drjit-core/half.h#L58). Our half has no half-vs-half comparison, so it converts both values to float and compares there anyway.

wjakob:
Wouldn't it be cleaner to override `std::numeric_limits` or use a custom wrapper?

wjakob:
Thanks!

PR #202, opened 2026-06-27
Set ORCv2 COFF responsibility flags on Windows

jeongseok-meta:
## Motivation

Dr.Jit uses a custom ORCv2 object-linking layer so it can keep its MCJIT-style memory manager. On Windows/COFF, generated constants such as `__real@...` and `__ymm@...` may not be covered by the symbols passed through `MaterializationResponsibility`. LLVM LLJIT enables responsibility override/auto-claim flags for its default COFF object layer, but the C API helper used here does not expose setters for those flags.

This came up while enabling the LLVM backend for the conda-forge Windows package: https://github.com/conda-forge/drjit-cpp-feedstock/pull/47

## Changes

- Add a Windows-only C++ bridge that casts the C API object layer back to `RTDyldObjectLinkingLayer` and enables the two COFF responsibility flags.
- Call it after creating Dr.Jit's custom object layer, preserving the existing memory manager.
- Add the LLVM include path for Windows dynamic-LLVM builds so the small bridge can include the ORC header.

## Testing

- `git diff --check`
- The equivalent downstream patch passed conda-forge Windows LLVM package CI in https://github.com/conda-forge/drjit-cpp-feedstock/actions/runs/28273630283 and was merged in https://github.com/conda-forge/drjit-cpp-feedstock/pull/47

I do not have a local Windows LLVM toolchain in this environment, so the Windows validation is from the conda-forge package CI above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,10 @@ add_library(
 target_include_directories(drjit-core PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 target_include_directories(drjit-core PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
+if (WIN32)
+  target_sources(drjit-core PRIVATE src/llvm_coff.cpp)
+endif()
+
 target_compile_features(drjit-core PUBLIC cxx_std_17)
 
 target_include_directories(drjit-core PRIVATE
@@ -325,6 +329,10 @@ endif()
 
 if (DRJIT_DYNAMIC_LLVM)
   target_compile_definitions(drjit-core PRIVATE -DDRJIT_DYNAMIC_LLVM=1)
+  if (WIN32)
+    find_package(LLVM REQUIRED)
+    target_include_directories(drjit-core PRIVATE ${LLVM_INCLUDE_DIRS})
+  endif()
 else()
   find_package(LLVM REQUIRED)
   target_include_directories(drjit-core PRIVATE ${LLVM_INCLUDE_DIRS})
--- a/src/llvm_coff.cpp
+++ b/src/llvm_coff.cpp
@@ -1,0 +1,16 @@
+#include <llvm/ExecutionEngine/Orc/RTDyldObjectLinkingLayer.h>
+
+extern "C" void jitc_llvm_set_coff_object_layer_flags(void *layer) {
+    /*
+       COFF generated constants such as __real@... and __ymm@... are not
+       covered by the MaterializationResponsibility symbols that Dr.Jit passes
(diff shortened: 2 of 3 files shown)


wjakob:
This kind of change requires linking against LLVM, correct? (You seem to be using symbols from the C++ lib).

wjakob:
Could you provide a reproducer that causes whatever issue motivated this change?

wjakob:
@jeongseok-meta You worked around my questions by closing the PR and adding a patch to the feedstock. Could I still ask for clarification and a reproducer? I am interested in tracking this down and fixing it.

jeongseok-meta:
Sorry for not following up earlier. You were right to ask for the clarification and reproducer.

The small bridge needs the LLVM C++ header, but it does not add a link dependency on the LLVM C++ libraries in `DRJIT_DYNAMIC_LLVM` mode. Both calls used by the bridge, `setOverrideObjectFlagsWithResponsibilityFlags` and `setAutoClaimResponsibilityForObjectSymbols`, are inline setters in `RTDyldObjectLinkingLayer.h`. The downstream CMake change therefore adds only `LLVM_INCLUDE_DIRS`; it does not add `target_link_libraries`. Runtime loading remains the existing `LLVM-C.dll` path.

The reproducer is the Windows dynamic-LLVM consumer used by the feedstock:

- [consumer source](https://github.com/conda-forge/drjit-cpp-feedstock/blob/5b2a79846c26a285eed98f28af0de32bd71d6855/recipe/tests/jit_llvm.cpp)
- [CMake setup](https://github.com/conda-forge/drjit-cpp-feedstock/blob/5b2a79846c26a285eed98f28af0de32bd71d6855/recipe/tests/CMakeLists.txt)
- [pre-patch Windows build](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_build/results?buildId=1544793&view=logs&jobId=c8207443-e636-510c-486c-006ab10b5e0e)
- [patched Windows build](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_build/results?buildId=1544883&view=logs&jobId=c8207443-e636-510c-486c-006ab10b5e0e)

With an unpatched Windows build using dynamic LLVM 21, `jit_init()` found LLVM successfully, but evaluating the generated LLVM array calculation failed while materializing COFF constants such as `__real@...` and `__ymm@...`. Enabling the two responsibility flags on the same custom object layer made that consumer pass while retaining Dr.Jit's memory manager.

I closed the PR too early instead of answering this. If this still looks like the right direction, I am happy to reopen/rework it and add a smaller upstream regression test if you have a preferred test location.

PR #204, opened 2026-06-27
Fix invalid mov.b8 PTX for 8-bit types in CUDA control flow

lnuic:
`UInt8`/`Int8` have no `mov.b8`, so CUDA symbolic loops/conditionals/calls carrying them aborted in ptxas.
Added small `jitc_cuda_mov`/`jitc_cuda_mov_zero` helpers for repetitive part, happy to drop or rename.


wjakob:
Thanks!

PR #205, opened 2026-07-10
Fix Metal intersection tags, float bitwise NOT, and symbolic gather guard

tjueterb:
Hi all,

this PR contains three fixes for drjit-core bugs fable found while updating my mitsuba fork to 3.9. The missing world_space_data tag is the most significant one, since it caused intersection misses on primitive shapes (e.g. sphere emitters) for me. Happy to split this into multiple PRs if you prefer.

## Metal codegen: missing world_space_data tag for custom intersectors

The custom-primitive intersection functions (sphere/disk/cylinder/ellipsoid, see resources/metal/) are declared with the MSL `world_space_data` intersection tag, since they operate on world-space rays. However, the codegen never included that tag in the `intersection_function_table<>` resource-handle type or in the `intersector<>` declaration at the trace-ray call site.

Per the MSL spec, an intersection function may only be used with a table/intersector whose tag set is a superset of the function's tags — otherwise the behavior is undefined. In practice, the mismatch meant the functions received object-space ray data where they expected world-space values: a coordinate-frame mismatch that produced silent intersection misses rather than a compile or validation error.

Fixed by adding `raytracing::world_space_data` to both sites: unconditionally in the table's resource-handle type (which is only emitted when an IFT is bound at all), and gated on `has_ift_local` at the intersector declaration, so triangle-only scenes are unaffected. The SDF-grid intersection function correctly omits the tag (it works on the object-space ray) and remains valid under a table with a superset of its tags.

## Metal codegen: bitwise NOT on floats

MSL has no `~` operator for floating-point types, but `VarKind::Not` emitted it unconditionally, producing invalid MSL whenever a non-Bool, non-integer variable needed a bitwise complement. Fixed by reinterpreting to the unsigned-int bit pattern, complementing, and reinterpreting back — mirroring the And/Xor cases directly above and the LLVM backend's handling.

## jit_var_gather: reject symbolic index/mask outside a recording session

`jit_var_scatter` already guards against symbolic inputs outside a symbolic recording session; `jit_var_gather` didn't. This allowed a gather to reference a symbolic loop-internal variable after its recording session closed (in my case: a differentiable scatter into an object outside a loop, replayed via a global AD traverse). On LLVM this aborts the process in `LLVMVerifyModule` ("Instruction does not dominate all uses!"). Fixed by adding the same guard scatter has, so it's now a clean RuntimeError everywhere. On CUDA the same operation did not abort; I did not verify whether the CUDA gradients were correct.

### DIFF
--- a/src/metal_eval.cpp
+++ b/src/metal_eval.cpp
@@ -631,6 +631,9 @@ static void jitc_metal_render(Variable *v) {
             Variable *a = jitc_var(v->dep[0]);
             if ((VarType) v->type == VarType::Bool)
                 fmt("$t $v = !$v;\n", v, v, a);
+            else if (jitc_is_float(v))
+                fmt("$t $v = as_type<$t>(($b)(~as_type<$b>($v)));\n",
+                    v, v, v, v, v, a);
             else
                 fmt("$t $v = ~$v;\n", v, v, a);
             break;
--- a/src/metal_eval.cpp
+++ b/src/metal_eval.cpp
@@ -131,8 +131,8 @@ static bool jitc_metal_render_resource_handle(const Variable *v,
             } else {
                 bool curves = scene && (scene->geometry_types_mask & 0x4u) != 0;
                 tname = curves
-                    ? "raytracing::intersection_function_table<raytracing::triangle_data, raytracing::instancing, raytracing::curve_data>"
-                    : "raytracing::intersection_function_table<raytracing::triangle_data, raytracing::instancing>";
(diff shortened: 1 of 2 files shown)


wjakob:
Hi Tobias - what's the reason for closing the PR?

tjueterb:
Hi Wenzel,

I noticed that the newer commits on mitsuba3/master went in the opposite route regarding the world_space_data tag compared to my commit. I think the other two fixes still apply, I will check asap and then reopen / open a new PR.

tjueterb:
I opened a new PR #206 and also an issue in the mitsuba repo that reproduces the `~float` error https://github.com/mitsuba-renderer/mitsuba3/issues/1915.

PR #207, opened 2026-07-17
Update ptr to variable map in call upload

dvicini:
It seems like the call upload function updated the variable's data pointer without updating the ptr_to_variable map, leaving the ptr_to_variable map in an outdated state (which in turn produced warnings during shutdown).

This PR adds the missing update.


wjakob:
Thanks!

PR #209, opened 2026-08-08
Add missing XXH3 state init

dvicini:
The `XXH3_state_t` object needs to be explicitly initialized to 0, otherwise we run into possibly undefined behavior due to the use of uninitialized memory (see also https://xxhash.com/doc/v0.8.2/xxhash_8h.html#afc79de39f0cf7e8d395f131c0d9d568e). The macro appears to not be included in the version of xxhash we include, but we can directly set the same fields to zero:

```cpp
xxh_state.seed = 0;
xxh_state.extSecret = nullptr;
```

### DIFF
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -123,6 +123,8 @@ static XXH128_hash_t jitc_cache_checksum(const CacheFileHeader &header,
                                          uint32_t source_size,
                                          const char *compressed) {
     XXH3_state_t xxh_state;
+    xxh_state.seed = 0;
+    xxh_state.extSecret = nullptr;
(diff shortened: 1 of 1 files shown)


wjakob:
Hi Delio -- if it needs to be zero-initialized, maybe just `{}` would do the job?

dvicini:
Nevermind - this problem is only relevant when using more recent xxhash versions (>= 0.8.0), not when using the one that is packaged with lz4 here. This will only matter if this version is ever updated.

The current version of the code already 0 initializes. This only changes in later versions. Zero-initializing now would effectively double-intialize, just to guard against futured updated, which makes no sense.

PR #213, opened 2026-09-12
OptiX: Adjust traversal depth to support motion transforms

dvicini:
Introducing motion blur makes the pipeline three levels deep: IAS -> Motion transform -> GAS.  This means it needs a traversal depth of 3 instead of 2.


wjakob:
I'd argue we definitely do not want to pay for anything if the user isn't using motion transforms. That includes this change here.

dvicini:
Yes I agree, sorry for not explaining more clearly in the PR description: If I understand the OptiX setup correctly, that should still hold. Mitsuba by default will use `OPTIX_TRAVERSABLE_GRAPH_FLAG_ALLOW_SINGLE_LEVEL_INSTANCING`, see https://github.com/mitsuba-renderer/mitsuba3/blob/master/src/render/scene_optix.inl#L127. 

For motion blur, we need to set `OPTIX_TRAVERSABLE_GRAPH_FLAG_ALLOW_ANY`.

PR #214, opened 2026-09-23
Fix undefined behaviour when initializing Optix program descriptors

matttsss:
## Summary
Fixes a segmentation fault inside `optixProgramGroupCreate` when compiling OptiX intersection program groups (`callable_units[i - 1].isect`) under **Clang** with `-ftrivial-auto-var-init=pattern`.

## Root Cause
`OptixProgramGroupDesc` contains an anonymous union whose members have different sizes:
- `raygen`, `miss`, `exception`, `callables`: 2 pointers (`16` bytes on 64-bit)
- `hitgroup`: 6 pointers (`48` bytes on 64-bit: `moduleCH`, `entryFunctionNameCH`, `moduleAH`, `entryFunctionNameAH`, `moduleIS`, `entryFunctionNameIS`)

Per C++ aggregate initialization rules (`[dcl.init.aggr]`), brace-initializing `OptixProgramGroupDesc pgd { };` only zero-initializes the **first** union member (`raygen`, i.e. the first 16 bytes), leaving the remaining 32 bytes of the union indeterminate.

When built with **Clang** and `-ftrivial-auto-var-init=pattern` (used to catch un-initialised memory reads), the uninitialized tail of the union is pre-filled with `0xAA`. Because `{}` only zeroes the first 16 bytes (`moduleCH` and `entryFunctionNameCH`), the middle 16 bytes (`pgd.hitgroup.moduleAH` and `pgd.hitgroup.entryFunctionNameAH`) remain `0xaaaaaaaaaaaaaaaa`. When `optixProgramGroupCreate` inspects `moduleAH`, it sees a non-null pointer, dereferences `0xaaaaaaaaaaaaaaaa`, and crashes with `SIGSEGV`.

<details>
<summary>Mitsuba reproducer</summary>
Compile with 

```
cmake -B build -GNinja \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DCMAKE_CXX_FLAGS="-ftrivial-auto-var-init=pattern"
```
and run 

```python
import drjit as dr
import mitsuba as mi

mi.set_variant("cuda_ad_rgb")

scene = mi.cornell_box()
scene['sphere'] = {
    'type': 'sphere',
    'bsdf': {
        'type': 'dielectric',
    }
}
scene = mi.load_dict(scene)

img = mi.render(scene, spp=1024, seed=0)
dr.eval(img)
```
</details>

## Fix

This issue is fixed by zero-initializing the struct as was done [here](https://github.com/mitsuba-renderer/drjit-core/blob/ea9f14402a4bdf1d9319931bec87429d1a879ef1/src/optix_core.cpp#L443).

The bug was tracked down with an LLM but the issue and fix were confirmed manualy.


wjakob:
Thanks!

